PR #5, opened 2016-08-30
Update Cloud SQL example with doc tags.

jeffmendoza:


piaxc:
Readme comments:
Replace link for "Google Cloud SQL" with https://cloud.google.com/sql/docs/

Replace
See the Cloud SQL access control page for more information.
(linking to access control page)
with
See the Cloud SQL Proxy documentation for more information.
and link  to:
https://cloud.google.com/sql/docs/mysql-connect-proxy#service-account

We tell them they need to have a service account already, so I think you could remove this:

First, create and download the JSON Service Account credentials following these steps. 

Since you changed the config file to use the connection string, need to update it in the readme also:

Next, open the cloudsql_deployment.yaml file in this repository and replace the values $PROJECT, REGION and $INSTANCE with the correct values for your SQL instance.

jeffmendoza:
@piaxc Since these comments don't affect the doc tags, or the contents between them. I'll merge this to unblock you and start a new PR with updates to the README.

PR #7, opened 2016-11-07
Fix missing cloudsql empty directory.

edwardfung123:
Missing directory cloudsql which should be a`emptyDir`. As a result, the pod cannot run.


edwardfung123:
I signed it!

jeffmendoza:
Thanks for the fix!

PR #15, opened 2017-03-29
hellonode: Handle SIGINT, add hostname and \n to response

ahmetb:
- Ctrl+C on a Docker container sends SIGINT, which is not handled
  properly by default (see nodejs/node#4182). Added a signal handler.
- Adding trailing newline to response for curl-friendliness.
- Added log message indicating server is starting.
- Added hostname to indicate that the container.

Signed-off-by: Ahmet Alp Balkan <ahmetb@google.com>

### DIFF
--- a/hellonode/server.js
+++ b/hellonode/server.js
@@ -15,11 +15,19 @@
  */
 
 var http = require('http');
+var os = require('os');
+
+process.on('SIGINT', function() {
+  console.log('shutting down...');
+  process.exit();
+});
 
 var handleRequest = function(request, response) {
   console.log('Received request for URL: ' + request.url);
   response.writeHead(200);
-  response.end('Hello, World!');
+  response.end('Hello, World!\nHostname: ' + os.hostname() + '\n');
 };
+
 var www = http.createServer(handleRequest);
+console.log('server listening on port 8080');
 www.listen(8080);
--- a/hellonode/server.js
+++ b/hellonode/server.js
@@ -14,20 +14,23 @@
  * limitations under the License.
  */
 
-var http = require('http');
-var os = require('os');
+const http = require('http');
+const os = require('os');
+
+const port = process.env.PORT || 8080;
 
 process.on('SIGINT', function() {
   console.log('shutting down...');
-  process.exit();
+  process.exit(1);
 });
 
 var handleRequest = function(request, response) {
-  console.log('Received request for URL: ' + request.url);
+  console.log(`Received request for URL: ${request.url}`);
   response.writeHead(200);
-  response.end('Hello, World!\nHostname: ' + os.hostname() + '\n');
+  response.end(`Hello, World!\nHostname: ${os.hostname()}\n`);
 };
 
 var www = http.createServer(handleRequest);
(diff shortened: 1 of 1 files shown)


hellonode/server.js
```diff
@@ -15,11 +15,19 @@
...
+process.on('SIGINT', function() {
+  console.log('shutting down...');
+  process.exit();
+});
 
 var handleRequest = function(request, response) {
   console.log('Received request for URL: ' + request.url);
   response.writeHead(200);
-  response.end('Hello, World!');
+  response.end('Hello, World!\nHostname: ' + os.hostname() + '\n');
```
MylesBorins: you could use a template string here
```js
`Hello, World!\nHostname: ${os.hostname()}\n`
```
you could also potentially cache the hostname above to avoid the function calls on each request, that is perhaps overkill though

hellonode/server.js
```diff
@@ -15,11 +15,19 @@
  */
 
 var http = require('http');
```
MylesBorins: modern node is using `const` rather than var for requires

hellonode/server.js
```diff
@@ -15,11 +15,19 @@
...
+var os = require('os');
+
```
MylesBorins: it might be nice to declare the port here, and give an option to declare it via env_var

```js
const port = process.env.NODE_PORT || 8080;
```

hellonode/server.js
```diff
@@ -15,11 +15,19 @@
...
+process.on('SIGINT', function() {
+  console.log('shutting down...');
+  process.exit();
```
MylesBorins: if you SIGINT should the process exit cleanly?

hellonode/server.js
```diff
@@ -15,11 +15,19 @@
...
+});
 
 var handleRequest = function(request, response) {
   console.log('Received request for URL: ' + request.url);
```
MylesBorins: can be done with template string

```js
console.log(`Received request for URL: ${request.url}`);
```

hellonode/server.js
```diff
@@ -15,11 +15,19 @@
...
   response.writeHead(200);
-  response.end('Hello, World!');
+  response.end('Hello, World!\nHostname: ' + os.hostname() + '\n');
 };
+
 var www = http.createServer(handleRequest);
+console.log('server listening on port 8080');
```
MylesBorins: you might want to put this message inside of a callback to the listen call, to ensure it doesn't print until the server is listening
```js
www.listen(port, ( ) => {
  console.log(`server is listening on port ${port}`);
});

File: hellonode/server.js [hunk 7]
```diff
@@ -14,12 +14,23 @@
  * limitations under the License.
  */
 
-var http = require('http');
+const http = require('http');
+const os = require('os');
+
+const port = process.env.PORT || 8080;
```
MylesBorins: nit: do you want to make this more unique than PORT? possible to already exist on system.

ahmetb: @MylesBorins kind of a convention, so I'll keep it.

PR #18, opened 2017-04-10
Import manifests for persistent-disk tutorial

ahmetb:
This tutorial's contents used to be served in a .zip on cloud.google.com,
now moving them here and updating. Major changes:

- use `Service` to do discovery instead of docker-links style env vars
- use Deployment instead of `Pod` deployment
- extract mysql root password to a `Secret` instead of embedding in YAML

Contact me for internal link to review the new tutorial if necessary.

Signed-off-by: Ahmet Alp Balkan <ahmetb@google.com>

### DIFF
--- a/wordpress-persistent-disks/mysql-service.yaml
+++ b/wordpress-persistent-disks/mysql-service.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: mysql
+  labels:
+    app: mysql
+spec:
+  type: ClusterIP
+  ports:
+    - port: 3306
+  selector:
+    app: mysql
--- a/wordpress-persistent-disks/mysql.yaml
+++ b/wordpress-persistent-disks/mysql.yaml
@@ -1,0 +1,36 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: mysql
+  labels:
+    app: mysql
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: mysql
+  template:
+    metadata:
+      labels:
+        app: mysql
+    spec:
+      containers:
+        - image: mysql:5.6
+          name: mysql
+          env:
+            - name: MYSQL_ROOT_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: mysql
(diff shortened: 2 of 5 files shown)


ahmetb:
@waprin can you please review this one?

wordpress-persistent-disks/mysql.yaml
```diff
@@ -0,0 +1,36 @@
...
+    spec:
+      containers:
+        - image: mysql:5.6
+          name: mysql
+          env:
+            - name: MYSQL_ROOT_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: mysql
+                  key: password
```
waprin: where does this secret come from?

ahmetb: @waprin it's created in the tutorial with `kubectl create secret`.

wordpress-persistent-disks/mysql-service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: mysql
+  labels:
+    app: mysql
+spec:
+  type: ClusterIP
```
waprin: We usually just let it default to this? Makes sense and is cleaner to me but maybe being explicit is better, idk.

ahmetb: yeah I wanted to make it explicit, as the other wordpress-service.yaml has type:LoadBalancer.

wordpress-persistent-disks/mysql.yaml
waprin: I don't see that in the tutorial linked...

waprin:
LGTM

wordpress-persistent-disks/mysql.yaml
mike-trewartha: agreed with @waprin - the process to create a secret is not in the tutorial.

ahmetb: @mike-trewartha for now please follow the https://cloud.google.com/container-engine/docs/tutorials/persistent-disk/. The new version of the tutorial (which this PR adds) is not published yet.

PR #20, opened 2017-05-21
update service name to avoid conflict

jankapusta:
If metadata name is same as app selector it causes pod to keep crashing
see https://github.com/kubernetes/kubernetes/issues/16906


jankapusta:
I signed it!

ahmetb:
@jankapusta are you actively running into this issue? While developing this sample, I did not have any problems on a Kubernetes v1.5 cluster.

ahmetb:
@jankapusta friendly ping. I am not able to reproduce this and planning to close. Let me know if you are able to reproduce this issue.

PR #22, opened 2017-06-20
Cloud Pub/Sub example

ahmetb:
Source code and the Dockerfile for the new Container Engine tutorial "Authenticating to Cloud Pub/Sub with Service Account".

### DIFF
--- a/cloud-pubsub/Dockerfile
+++ b/cloud-pubsub/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7-slim
+FROM python:3-slim
 COPY . .
 RUN pip install -r requirements.txt
 ENTRYPOINT ["python", "-u", "main.py"]
--- a/cloud-pubsub/Dockerfile
+++ b/cloud-pubsub/Dockerfile
@@ -1,0 +1,4 @@
+FROM python:2.7-slim
+COPY . .
+RUN pip install -r requirements.txt
+ENTRYPOINT ["python", "-u", "main.py"]
--- a/cloud-pubsub/main.py
+++ b/cloud-pubsub/main.py
@@ -6,9 +6,8 @@ prints to standard output.
 """
 
 import datetime
-from google.cloud import pubsub
-from google.cloud.pubsub.subscription import AutoAck
 
+from google.cloud import pubsub
 
 PUBSUB_TOPIC = 'eco'
 PUBSUB_SUBSCRIPTION = 'echo-read'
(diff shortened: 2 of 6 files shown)


cloud-pubsub/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:2.7-slim
```
jonparrott: It's 2017, please use Python 3.

cloud-pubsub/main.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+"""Pub/Sub pull example on Google Container Engine.
+
+This program pulls messages from a Cloud Pub/Sub topic and
+prints to standard output.
+"""
+
+import datetime
```
jonparrott: imports are grouped as

```
standard library

third party

local
```

So this should be:

```python
import datetime

from google.cloud import pubsub
from google.cloud.pubsub.subscription import AutoAck
```

cloud-pubsub/main.py
```diff
@@ -0,0 +1,32 @@
...
+from google.cloud import pubsub
+from google.cloud.pubsub.subscription import AutoAck
```
jonparrott: our style guide strongly prefers import modules instead of importing the members inside of them.

cloud-pubsub/main.py
```diff
@@ -0,0 +1,32 @@
...
+PUBSUB_TOPIC = 'eco'
+PUBSUB_SUBSCRIPTION = 'echo-read'
+
+
+def main():
+    """Continuously pull messages from subsciption"""
+    client = pubsub.Client()
+    subscription = client.topic(PUBSUB_TOPIC).subscription(PUBSUB_SUBSCRIPTION)
+
+    print 'Pulling messages from Pub/Sub subscription...'
```
jonparrott: print is a function, so use `print(...)`.

cloud-pubsub/main.py
```diff
@@ -0,0 +1,32 @@
...
+    while True:
+        with AutoAck(subscription, max_messages=10) as ack:
```
jonparrott: Just so you know - this client library is going to change significantly within the next month.

cloud-pubsub/main.py
```diff
@@ -0,0 +1,32 @@
...
+            for _, message in list(ack.items()):
+                msg = "[{0}] ID={1} Data={2}"
+                print msg.format(datetime.datetime.now(),
```
jonparrott: for simple format strings, no need to assign them to a temporary variable.

cloud-pubsub/main.py
ahmetb: that sounds fine as we lock to a version today.

ahmetb:
Addressed feedback. PTAL again @jonparrott.

PR #28, opened 2017-09-07
Updated hello-app/Dockerfile to be more friendly to older docker versions

kurtisvg:
@lesv 

This removes the multi-staging from the Dockerfile, which requires Docker 1.17.05 or higher. G-Cloud webshell runs a lower version than this. 

Previously the docker compiled the app in one stage and then copied into another one for minimization. This change just builds and hosts the app in the same container since multi-staging isn't possible.

### DIFF
--- a/hello-app/Dockerfile
+++ b/hello-app/Dockerfile
@@ -1,8 +1,6 @@
 FROM golang:1.8-alpine
-ADD . /go/src/hello-app
-RUN go install hello-app
-
-FROM alpine:latest
-COPY --from=0 /go/bin/hello-app .
+ADD . /hello-app
+WORKDIR /hello-app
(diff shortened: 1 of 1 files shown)


lesv:
This makes it work on older versions of Docker - like gcloud web shell

ahmetb:
It would be nice if I reviewed this as well. This will make the quickstart noticably slower if we reupload this version of this image to GCR. The image size is also why we rewrote this in Go instead of Node.

Is there a tutorial that currently builds this Dockerfile on Google Cloud Shell?

kurtisvg:
The tutorial is found here: https://cloud.google.com/container-engine/docs/tutorials/hello-app

PR #30, opened 2017-10-04
Update cloudsql_deployment.yaml

piaxc:
Apparently we've had several users lately who did not realize they needed to remove the square brackets around INSTANCE_CONNECTION_NAME. This could be because of the (required) square brackets around the entire command. Switched to angle brackets instead. Also, updated proxy version to most recent.

### DIFF
--- a/cloudsql/cloudsql_deployment.yaml
+++ b/cloudsql/cloudsql_deployment.yaml
@@ -15,9 +15,9 @@ spec:
           env:
             - name: WORDPRESS_DB_HOST
               # Connect to the SQL proxy over the local network on a fixed port.
-              # Change the [PORT] to the port number used by your database
-              # (e.g. 3306).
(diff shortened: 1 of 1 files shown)


ahmetb:
We use this pattern in many other places/products. I'm not certain if this is going to cause any less confusion.

ahmetb:
I wish our code import script supported extra highlighting of certain keywords, in that case we would not need to do `<PORT>` or `[PORT]`.

jeffmendoza:
I think we tend towards `<> `

jeffmendoza:
> I wish our code import script supported extra highlighting of certain keywords, in that case we would not need to do `<PORT>` or `[PORT]`.

Yeah, but people still need to see it in their editor.

piaxc:
Page published. Thanks.

PR #31, opened 2017-10-10
Add 0.0.0.0 to proxy connection string

piaxc:
As per pull request here (actually only a comment addition):
https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/127

I believe we need the 0.0.0.0 in this file also.

### DIFF
--- a/cloudsql/cloudsql_deployment.yaml
+++ b/cloudsql/cloudsql_deployment.yaml
@@ -43,7 +43,7 @@ spec:
         - image: gcr.io/cloudsql-docker/gce-proxy:1.11
           name: cloudsql-proxy
           command: ["/cloud_sql_proxy", "--dir=/cloudsql",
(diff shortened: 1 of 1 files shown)


ahmetb:
@piaxc can you verify this (and https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/127) with Cloud SQL engineering. Binding to 0.0.0.0 (all interfaces) instead of localhost (loopback interface) can cause security issues through accidental exposure.

ahmetb:
It's probably fine, as proxy container has no `ports` section exposed outside the Pod but it would be still good to check. I'm curious why this is suddenly needed —it's been working fine.

ahmetb:
I think this change is not needed. I'm able to use the existing cloudsql_deployment.yaml file and run `kubectl expose deployment wordpress --type=LoadBalancer` to get a WordPress set up that's working.

PR #32, opened 2017-10-16
Clean up cloudsql example, bring it up to date

ahmetb:
This makes the Cloud SQL example coherent with the updated documentation at
https://cloud.google.com/sql/docs/mysql/connect-container-engine.

Main updates:
- Hardcode port 3306 because wordpress doesn't work with postgresql.
- Update README.md to NOT to include instructions but refer to the cloud sql
  docs for setup.
- Remove unnecessary parts from the manifest.

cc: @jeffmendoza @piaxc


piaxc:
If Wordpress doesn't work with PostgreSQL, we need to find something that does. The sample is presented for both database engines.

ahmetb:
@piaxc Unless Cloud SQL magically detects and supports both Postgres and MySQL wire formats at the same time, it means this never worked in the first place so it should not block this pull request.

ahmetb:
That said I realized the YAML snippets are also coming from the cloudsql_deployment.yaml file, so we can't really hardcode the ports.

If Postgres never worked on MySQL I think we might need to duplicate the file so we have:
- a manifest for MySQL (so people don't need to change `<PORT>` to deploy)
- and one for PostgreSQL (that hardcodes pgsql ports, but requires user to specify an `<IMAGE>`, it's not for wordpress but this file is used to embed snippets in the docs).

davisshaver:
+1 to merging this PR.

ahmetb:
@piaxc @jeffmendoza I just added a patch that separates mysql and postgres into separate files. MySQL one will continue to work as WordPress, and PostgreSQL manifest contains an `image: <CONTAINER-IMAGE>` placeholder.

I'm also preparing the internal patch to accommodate this patch.

ahmetb:
@jeffmendoza updated the PR.

jeffmendoza:
@ahmetb Lobsters might be a good app to use postgres ( or both? ) I have an example [here](https://github.com/GoogleCloudPlatform/kubernetes-workshops/tree/master/state). That one uses mysql, but you can change out the database adapter.

ahmetb:
I suspect Lobsters has more moving parts/components than WordPress, which might not be understood easily. Even with WordPress the fact that user has to create a Load Balancer and move on to setting up the blog is not very ideal just to see if the DB connection is working.

I was thinking of building a super simple container image that connects to the database and lists some database info (tables/users etc) to stdout.

But I think right now we have decent example coverage around this to get people up and running so I'm suspecting it might not require more investment currently.

jeffmendoza:
I agree with removing the broken suggestion (wordpress + postgress) immediately. However, leaving it as `<CONTAINER-IMAGE>` isn't a good long or medium term solution.

Any container image that does something would be good soon. A super simple one as you described would be good. 

I just remembered that Lobsters requires a db create/seed setup step, so it is not ideal.

ahmetb:
I think the outstanding internal CL will pick up these changes and point people to the `postgresql_deployment.yaml` where they can see an example of Cloud SQL Proxy sidecar. It's good that we now have something for Postgres even though it is not an e2e example.

jeffmendoza:
:+1:

PR #42, opened 2018-01-12
cloudsql: remove --dir=/cloudsql from samples

ahmetb:
There was an unused --dir=/cloudsql directive passed to cloudsql image and
the resulted unix socket was not used since the sample makes use of TCP
connection between the app and the cloudsql sidecar container.

cc: @athenashi @piaxc


ahmetb:
Merging directly due to confirmation from @athenashi and manual verification completed by me.

PR #48, opened 2018-03-09
import of hello-app-iap

danisla:


hello-app-iap/README.md
```diff
@@ -0,0 +1,20 @@
+# ESP Sample App
```
ahmetb: what's ESP

hello-app-iap/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Displays the email address of the logged in user from the value of the `x-goog-authenticated-user-email` header inserted by the Identity-Aware Proxy.
+
+## Building
```
ahmetb: Delete this and just say "this image is available at `gcr.io/google-samples/hello-app-iap:lastest`

hello-app-iap/app.py
```diff
@@ -0,0 +1,45 @@
...
+import os
+PORT = int(os.environ.get("PORT", "8080"))
+class RequestHandler(BaseHTTPRequestHandler):
+    def do_GET(self):
+        self.send_response(200)
+        self.send_header('Content-type','text/html')
+        self.end_headers()
+        self.wfile.write(bytes("""
+        <!doctype html><html>
+    <head><title>ESP Sample App</title><link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/0.100.2/css/materialize.min.css"></head>
```
ahmetb: yeah all this materalize stuff is really not necessary, you can just do

```html
<h1>Hello %s</h1>
<ul>
<li> link 1 </li>
<li> link 2 </li>
</ul>

File: hello-app-iap/app.py [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+        <div class="card-content white-text">
+            <h4>Hello %s</h4>
+        </div>
+        <div class="card-action">
+            <a href="/_gcp_iap/identity">Identity JSON</a>
+            <a href="/_gcp_iap/clear_login_cookie">Logout</a>
+        </div>
+    </div></div></div></div>
+    </body></html>
+        """ % self.headers.get("x-goog-authenticated-user-email","unauthenticated user").split(':')[-1], "utf8"))
```
ahmetb: the logic on this line is too complicated, extract it  into a variable.

hello-app-iap/app.py
```diff
@@ -0,0 +1,45 @@
...
+print("Listing on port", PORT)
```
ahmetb: listening

hello-app-iap/cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
+steps:
```
ahmetb: no need for this file for 2 reasons:
1. I'll be pushing it
2. you can do the same thing without this file like `gcloud container builds submit -t TAG .`

hello-app-iap/deployment.yaml
```diff
@@ -0,0 +1,26 @@
...
+        app: hello-app-iap
+    spec:
+      containers:
+      - name: app
+        image: gcr.io/google-samples/hello-app-iap:1.0.0
+        ports:
+          - containerPort: 8080
+        readinessProbe:
+          httpGet:
+            host: "{{ENDPOINT}}"
```
ahmetb: do you need `host` in this case? also how does the probe authenticate to the `/_gcp_iap/identity` to get an `http 200 OK`?

hello-app-iap/deployment.yaml
```diff
@@ -0,0 +1,26 @@
...
+            path: /_gcp_iap/identity
+            port: 443
+            scheme: HTTPS
+          periodSeconds: 10
```
ahmetb: just roll with defaults on this readiness probe unless you have a good reason to customize.

hello-app-iap/deployment.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: hello-app-iap
+spec:
+  replicas: 1
```
ahmetb: omit this value, it's default

hello-app-iap/app.py
```diff
@@ -0,0 +1,45 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+ 
+ 
+from http.server import HTTPServer, BaseHTTPRequestHandler
+import os
+PORT = int(os.environ.get("PORT", "8080"))
+class RequestHandler(BaseHTTPRequestHandler):
+    def do_GET(self):
```
ahmetb: This app will work just fine without IAP and that's not good. I recommend you do these 2:

1. detect the header, if missing return a simple text response: `error: application is not running behind Identity-Aware Proxy (x-dsfldsjkfdlsfjkas header missing)`

2. when you find the header use the methods described at htps://cloud.google.com/iap/docs/signed-headers-howto#verify_the_id_token_header to validate the claim. we know that for proxies coming from IAP, these headers CANNOT be forged. However, what if you're on a GKE cluster and another pod within the cluster queries your NodePort service directly by forging the headers?

ahmetb:
Also since you're not deriving this from `hello-app`, we should not call it `hello-app-iap`, that just creates confusion. We can say `iap-example` or `iap-sample-app` or something.

hello-app-iap/deployment.yaml
danisla: It doesn’t. This path only exists when IAP is enabled on the backend service. This probe ensures that traffic never goes to the pod without IAP first being enabled.

hello-app-iap/app.py
danisla: This is what ESP is for, it does this JWT validation against the claim for you.

danisla: I think we should have multiple examples for IAP.
1. Pushing the responsibility of validating the JWT to the app.
2. Using ESP to validate the JWT (no user code change required)
3. Using Envoy to validate the JWT.

ahmetb: But how does ESP help you prevent other pods in the cluster circumvent the external LB and call the internal pod directly? Thats why you need in-service JWT validation too, I think.

I’m a bit against talking about stuff like envoy in our tutorials just yet.

ahmetb:
Closing this. Happy to revisit if/when we decide to to revisit an IAP sample.

PR #50, opened 2018-03-20
remove un-used cloudsql volume referenced in cloudsql proxy sample

rickypai:
there is a cloudsql volume that is defined and not mounted anywhere. it does not seem to be used at all.


ahmetb:
I thought I removed it in #42, I must have missed it. Thanks.

PR #52, opened 2018-04-06
[DO NOT MERGE] Updated Wordpress persistent disk tutorial to use PVCs

davidz627:
Wordpress persistent disk tutorial was out of date. This is an update to make it use PersistentVolumeClaims and Dynamic Provisioning.

DO NOT MERGE UNTIL INTERNAL DOCS CHANGE IS REVIEWED.


wordpress-persistent-disks/mysql-claim.yaml
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: mysql-claim
+spec:
+  accessModes:
+    - ReadWriteOnce
+  volumeMode: Filesystem
```
ahmetb: I don't think this field is necessary? If default, please omit. (ditto: below)

davidz627: resolved

ahmetb:
Ping me before submitting the internal changelist.

PR #66, opened 2018-05-11
Add HPA to Pub/Sub example

MaciekPytel:
This will be used in upcoming HPA external metrics tutorial. Adding sleep to simulate processing time is needed, so it's easy to generate load that a single instance can't handle and trigger autoscaling.

### DIFF
--- a/cloud-pubsub/main.py
+++ b/cloud-pubsub/main.py
@@ -6,6 +6,7 @@ prints to standard output.
 """
 
 import datetime
+import time
 
 from google.cloud import pubsub
 
@@ -22,9 +23,21 @@ def main():
     while True:
         with pubsub.subscription.AutoAck(subscription, max_messages=10) as ack:
             for _, message in list(ack.items()):
-                print("[{0}] ID={1} Data={2}".format(datetime.datetime.now(),
-                                                     message.message_id,
-                                                     message.data))
+                print("[{0}] Received message: ID={1} Data={2}".format(
+                    datetime.datetime.now(),
+                    message.message_id,
+                    message.data))
+                process(message)
+
+
+def process(message):
+    """Process received message"""
+    print("[{0}] Processing: {1}".format(datetime.datetime.now(),
+                                         message.message_id))
+    time.sleep(3)
+    print("[{0}] Processed: {1}".format(datetime.datetime.now(),
+                                        message.message_id))
+
 
 if __name__ == '__main__':
     main()
--- a/cloud-pubsub/pubsub-hpa.yaml
+++ b/cloud-pubsub/pubsub-hpa.yaml
@@ -1,0 +1,19 @@
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: pubsub
+spec:
+  minReplicas: 1
+  maxReplicas: 5
+  metrics:
+  - external:
+      metricName: pubsub.googleapis.com|subscription|num_undelivered_messages
+      metricSelector:
+        matchLabels:
+          resource.labels.subscription_id: echo-read
+      targetAverageValue: "2"
+    type: External
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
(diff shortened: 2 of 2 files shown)


cloud-pubsub/main.py
```diff
@@ -6,13 +6,23 @@
...
 
 from google.cloud import pubsub
 
 PUBSUB_TOPIC = 'eco'
 PUBSUB_SUBSCRIPTION = 'echo-read'
 
 
+def process(message):
+    """Process received message"""
+    print("[{0}] Processing: {1}".format(datetime.datetime.now(),
```
ahmetb: please use `message.message_id` in this method. text can be long or non-unique.

cloud-pubsub/main.py
```diff
@@ -6,13 +6,23 @@
...
 import datetime
+import time
 
 from google.cloud import pubsub
 
 PUBSUB_TOPIC = 'eco'
 PUBSUB_SUBSCRIPTION = 'echo-read'
 
 
+def process(message):
```
ahmetb: please consider moving below main() as that's the actual thing we want to show.

cloud-pubsub/hpa/hpa.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: autoscaling/v2beta1
```
ahmetb: can you rename hpa/ dir to manifests/

MaciekPytel: Should I also move deployment yaml to the new folder? That would break the existing tutorial between merging PR and updating the link in there.

ahmetb: in that case, let's move this to ../ and call it horizontalpodautoscaler.yaml

ahmetb: or pubsub-hpa.yaml.

cloud-pubsub/main.py
MaciekPytel: done

cloud-pubsub/main.py
MaciekPytel: done

cloud-pubsub/hpa/hpa.yaml
MaciekPytel: done

cloud-pubsub/main.py
```diff
@@ -22,9 +23,21 @@ def main():
     while True:
         with pubsub.subscription.AutoAck(subscription, max_messages=10) as ack:
```
ahmetb: @MaciekPytel have you tested this program locally? I say this because I think this might be querying a batch of 10 messages on every loop, so you need to send min 11 messages to see the delay.

we can change it to 1 if you need to. Please try it out and let me know.

cloud-pubsub/pubsub-hpa.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: pubsub
+  namespace: default
```
ahmetb: `namespace:`  is not necessary please avoid defaults

cloud-pubsub/main.py
MaciekPytel: Right, but it still sleeps 3 seconds per message, so it doesn't really matter (the throughput is the same). I wouldn't send a PR if I hadn't tested it locally :)

cloud-pubsub/pubsub-hpa.yaml
MaciekPytel: done

cloud-pubsub/main.py
MaciekPytel: For additional context - it takes over a minute to trigger autoscaling, I'm using a bash loop to send many more messages than 10.

ahmetb:
Just repushed gcr.io/google-samples/pubsub-sample:v1, PTAL.

PR #70, opened 2018-05-27
Improve custom metric example to handle SIGTERM

### LINKED ISSUE #69: SIGTERM not handled by direct-to-sd custom metrics example
Hi Guys/Girls! =)

I was trying to use the `custom-metrics-autoscaling/direct-to-sd/` example in my cluster but I notice that it is not handling SIGTERM (and other signals...) when we terminate a pod. I made this change in a fork and now it's seems working properly: https://github.com/alissonperez/kubernetes-engine-samples/commit/ee577dbdbbdee5f7a4a1083feb6e97651403f4f5

Please, take a look, if you agree I can open a PR with these changes!

Thank you!

alissonperez:
Using `/bin/sh -c` to execute `direct-to-sd` we can't handle properly process signals (like SIGTERM).

closes #69

### DIFF
--- a/custom-metrics-autoscaling/direct-to-sd/custom-metrics-sd.yaml
+++ b/custom-metrics-autoscaling/direct-to-sd/custom-metrics-sd.yaml
@@ -31,10 +31,8 @@ spec:
         run: custom-metric-sd
     spec:
       containers:
-      - command:
-        - /bin/sh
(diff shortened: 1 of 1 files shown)


ahmetb:
@aleksandra-malinowska FYI

PR #81, opened 2018-10-22
Update Deployment examples to v1 API

mistyhacks:
Deployment graduated to v1 in Kubernetes 1.9.

PR #82, opened 2018-11-05
Fix small errors of cloud-pubsub, wordpress-persistent-disks example made from #81 

lonyele:
As #81 upgrading the apiVersion, I got this error when `kubectl create -f pubsub.yaml`   --> `error validating data: ValidationError(Deployment.spec):ent.spec): missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec`

It seems like It has to explicitly match labels from this version. I just added the field. Also It's not a problem for now, but there is a typo on `main.py`. It seems like the image on gcr is built on top of fixed typo version.


lonyele:
I signed it!

PR #94, opened 2019-04-26
Add an example for Cloud IAP using the hello-app

tristan-lin:


hello-app-iap/README.md
```diff
@@ -0,0 +1,84 @@
...
+> **Note:** This application is a copy of [hello-app](../hello-app) sample.
+> See that directory for more details on this sample.
+
+In this demo, we open two ports on the hello-app container, 8081
+and 8082, for two services. For each service, we assign a path, and then 
+use Cloud IAP to configure different access for it. To serve HTTPS traffic, we 
+will create a Google managed SSL certificate for the endpoint, so make sure you are using a 1.12+
+cluster.
+
+## Build the image
```
ahmetb: Ideally we shouldn't document it here. This repo is just for the code backing the documentation.

There's an open bug (b/37857731) to create a tutorial explaining this. 
Do you mind moving this to a cloud.google.com tutorial in GKE doc set?

hello-app-iap/main.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func main() {
+	// use port1 serves service1 and port2 serves service2
+	const (
+		port1 = "8081"
+        	port2 = "8082"
+	)
+
+	finish := make(chan bool)
```
ahmetb: this doesn't do anything.

hello-app-iap/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"log"
+	"net/http"
+	"os"
+)
+
+func main() {
+	// use port1 serves service1 and port2 serves service2
+	const (
+		port1 = "8081"
+        	port2 = "8082"
```
ahmetb: this code isn't gofmtted. :)

hello-app-iap/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+	finish := make(chan bool)
+
+	server1 := http.NewServeMux()
+	server1.HandleFunc("/", hello1)
+
+    	server2 := http.NewServeMux()
+	server2.HandleFunc("/", hello2)
+
+   	go func() {
```
ahmetb: just do `go log.Fatal(http.ListenAndServe(...))` for simplicity here

hello-app-iap/main.go
```diff
@@ -0,0 +1,73 @@
...
+	    	err := http.ListenAndServe(":"+port2, server2)
+        	log.Fatal(err)
+	}()
+
+    	<-finish
+	
+}
+
+// hello1 prints "Hello world from service1"
+func hello1(w http.ResponseWriter, r *http.Request) {
```
ahmetb: hmm you can potentially merge hello1 hello2 into one method and just log:


`log.Printf("received request on address %s", req.Host)`

would that work?

hello-app-iap/main.go
tristan-lin: Done.

hello-app-iap/main.go
tristan-lin: I believe http.ListenAndServe() should live in a literal, or it won't work, at least not in my test bed.

hello-app-iap/main.go
tristan-lin: Yes it works. But I want to make this demo capable of multiple domains in the future. Or it can work for multiple domains with minor changes. So I am still prone to keep two ports here. Make sense?

ahmetb: I guess you can keep separate, it's ok.

hello-app-iap/main.go
ahmetb: Doesn't sound right.

You can do

```
go log.fatal(listenandserve(...))
go log.fatal(listenandserve(...))
<-ch //block forever
```

ahmetb: Also ,you don't have to fork both servers in the background, you can do one foreground one background:

```
go log.Fatalf("failed to listen on server1: %+v", srv1.ListenAndServe())
log.Fatalf("failed to listen on server2: %+v", srv2.ListenAndServe())
```

tristan-lin: Hi, 
This is what I have done in the latest commit. Looks good?
```
go func() {
  log.Fatal(http.ListenAndServe(":"+port1, server1))
}()
log.Fatal(http.ListenAndServe(":"+port2, server2))
```

tristan-lin:
@ahmetb 
Hi, I have reworded the README and removed most of the tutorials. Is that OK we move forward?
BTW, how could I build an image and upload to `gcr.io/google-samples/` like other `hello-app` samples?

Daniel-Sanche:
What's the target of this sample? Is a doc being written around this?

PR #96, opened 2019-05-16
Add wordpress_cloudsql.yaml.template

aminamansour:


kurtisvg:
Fixes #97

DanSanche:
The YAML content LGTM, although I'm not completely aware of the context behind this (I assume it's going to be added to the [existing wordpress tutorial?](https://cloud.google.com/kubernetes-engine/docs/tutorials/persistent-disk)).

I agree we should look into merging this with the cloudsql content at some point, but I'm not too sure about the best way to structure everything yet, so that decision can come later if this is time sensitive

Issue #92, opened 2019-03-23
[cloudsql] Update Cloud SQL Proxy to the latest version


kelvins:
Does it make sense to update the Cloud SQL Proxy in the examples to the latest [version 1.13](https://github.com/GoogleCloudPlatform/cloudsql-proxy/releases/tag/1.13)?
For example: [here](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/cloudsql/postgres_deployment.yaml#L44), [here](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/cloudsql/mysql_wordpress_deployment.yaml#L43) and [here](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/service-catalog/cloud-sql-mysql/manifests/user-deployment.yaml#L47).
I'm not sure if it breaks something but, since it is a minor version upgrade, I believe it is backward compatible.

PR #99, opened 2019-05-24
Update example with hint towards Private IP instructions. 

### LINKED ISSUE #80: Mention --ip_address_types Cloud SQL proxy flag
@kurtisvg should we mention the --ip_address_types flag in the deployment yamls:

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/cloudsql/postgres_deployment.yaml
https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/cloudsql/mysql_wordpress_deployment.yaml

Perhaps in the command args array, add something like:
```
# Uncomment below to force the pod to use the private IP address of the Cloud SQL instance.
# The instance must have a private IP on the same network and in the same region.
# "-ip_address_types=PRIVATE",
```

kurtisvg:
Fixes #80, #92

### DIFF
--- a/cloudsql/mysql_wordpress_deployment.yaml
+++ b/cloudsql/mysql_wordpress_deployment.yaml
@@ -40,7 +40,7 @@ spec:
         # $PROJECT:$REGION:$INSTANCE
         # [START proxy_container]
         - name: cloudsql-proxy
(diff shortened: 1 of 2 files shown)


ahmetb:
Lgtm. Fyi @DanSanche is the new maintainer of this repo.

PR #103, opened 2019-06-12
Support MySQL, see b/64102112

grant:
See: https://github.com/docker-library/mysql/issues/186
b/64102112

### DIFF
--- a/wordpress-persistent-disks/mysql.yaml
+++ b/wordpress-persistent-disks/mysql.yaml
@@ -15,7 +15,7 @@ spec:
         app: mysql
     spec:
       containers:
-        - image: mysql:5.6
+        - image: mysql:5.7
           name: mysql
           args:
             - "--ignore-db-dir=lost+found"
--- a/wordpress-persistent-disks/mysql.yaml
+++ b/wordpress-persistent-disks/mysql.yaml
@@ -17,6 +17,8 @@ spec:
       containers:
         - image: mysql:5.6
           name: mysql
+          args:
+            - "--ignore-db-dir=lost+found"
           env:
             - name: MYSQL_ROOT_PASSWORD
               valueFrom:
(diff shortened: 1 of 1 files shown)


ahmetb:
cc: @sanche is the new maintainer.

wordpress-persistent-disks/mysql.yaml
```diff
@@ -17,6 +17,8 @@ spec:
       containers:
         - image: mysql:5.6
           name: mysql
+          args:
+            - "--ignore-db-dir=lost+found"
```
ahmetb: this fix is only necessary if you're gonna change image to mysql:5.7.

this manifest actually works just fine with the currently listed image. the bug only happens when people try to use a newer image.

grant: OK. I've changed the image to 5.7.

ahmetb:
The bug  had some discussion on this. We can straight move to :8, if it works.

grant:
> cc: @sanche is the new maintainer.

I believe you mean sanche@ or @DanSanche.

---

@DanSanche PTAL and suggest changes or merge.

DanSanche:
I agree with Ahmet that if we're going to update MySQL, we should try to update straight to :8. Has anyone tried it to see if it works? Any known issues?

I can try to run through it myself, but it might be in a couple days

DanSanche:
I just tried out v8. It looks like lost+found isn't an issue, but we do need to change the auth plugin:

```
        - image: mysql:8
          name: mysql
          args:
            - "--default-authentication-plugin=mysql_native_password"
```

grant:
Hey @DanSanche, can you recommend a way for me to make progress on this PR?
LMK what you'd like to do.

DanSanche:
I think your last commit should be all we need. I'll run through it again today to verify

PR #104, opened 2019-06-25
"Deploying a language-specific app" quickstart code

bshaffer:
The following code is for the "Deploying a language-specific app" quickstart.


quickstart/java/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# It's important to use OpenJDK 8u191 or above that has container support enabled.
+# https://hub.docker.com/r/adoptopenjdk/openjdk8
+# https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
+FROM adoptopenjdk/openjdk8:jdk8u202-b08-alpine-slim
+
+# Copy the jar to the production image from the builder stage.
+COPY --from=builder /app/target/helloworld-*.jar /helloworld.jar
+
+# Run the web service on container startup.
+CMD ["java","-Djava.security.egd=file:/dev/./urandom","-Dserver.port=${PORT}","-jar","/helloworld.jar"]
```
kurtisvg: Is PORT supposed to be a variable? I don't see it declared.

bshaffer: It's declared  in  `deployment.yaml` as an environment variable

quickstart/nodejs/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "knative-serving-helloworld",
+  "version": "1.0.0",
+  "description": "Simple hello world sample in Node",
```
fhinkel: Can we leave out the word `simple`?

quickstart/go/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Use the offical Golang image to create a build artifact.
+# This is based on Debian and sets the GOPATH to /go.
+# https://hub.docker.com/_/golang
+FROM golang:1.12 as builder
+
+# Copy local code to the container image.
+WORKDIR /go/src/github.com/knative/docs/helloworld
```
tbpg: Update (usually `/app` or `/code` or something).

quickstart/go/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY . .
+
+# Build the command inside the container.
+# (You may fetch or manage dependencies here,
+# either manually or with a tool like "godep".)
```
tbpg: I'd remove the reference to downloading dependencies and `godep`. If we do reference dependencies, we should show copying just the `go.mod` and `go.sum` files, running `go mod download`, copying everything else, then doing `go build`.

cc @grayside

quickstart/go/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
+
+# Use a Docker multi-stage build to create a lean production image.
+# https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
+FROM alpine
+RUN apk add --no-cache ca-certificates
+
+# Copy the binary to the production image from the builder stage.
+COPY --from=builder /go/src/github.com/knative/docs/helloworld/helloworld /helloworld
```
tbpg: Update based on path from above.

quickstart/go/helloworld.go
```diff
@@ -0,0 +1,30 @@
...
+func handler(w http.ResponseWriter, r *http.Request) {
+	log.Print("Hello world received a request.")
+	target := os.Getenv("TARGET")
+	if target == "" {
+		target = "World"
+	}
+	fmt.Fprintf(w, "Hello %s!\n", target)
+}
+
+func main() {
```
tbpg: Please move the `main` to the top (above `handler`).

quickstart/go/helloworld.go
```diff
@@ -0,0 +1,30 @@
...
+	log.Print("Hello world sample started.")
```
tbpg: I'd delete this and do something like the following right before the call to `log.Fatal` below.

```
log.Printf("Listening on localhost:%s", port)
```

quickstart/ruby/app.rb
```diff
@@ -0,0 +1,8 @@
+require 'sinatra'
+
+set :bind, '0.0.0.0'
```
frankyn: I think you're missing `set :port, 8080` given the default port Sinatra listens on is 4567.

quickstart/python/app.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+from flask import Flask
+
+app = Flask(__name__)
+
+@app.route('/')
+def hello_world():
+    target = os.environ.get('TARGET', 'World')
```
engelke: I don't see how TARGET ever gets set, so is this needed?

quickstart/python/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# https://hub.docker.com/_/python
+FROM python:3.7
+
+# Copy local code to the container image.
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+COPY . .
+
+# Install production dependencies.
+RUN pip install Flask gunicorn
```
engelke: This is fine for this app, but the convention is to include a requirements.txt file with all dependencies (including versions if desired), and then run pip install -r requirements.txt instead of this.

quickstart/go/Dockerfile
bshaffer: sorry, I don't understand what this means

quickstart/python/app.py
bshaffer: No it isn't needed, but I've left it in all the examples for now just to make the code slightly more interesting.

quickstart/python/Dockerfile
bshaffer: The reason this was done is so that the users can copy-paste just two files (the `Dockerfile`  and `app.py`) to  get the  app running. What if I add a comment like  this?

```
# NOTE: This is for simplicity. Convention is to use a requirements.txt in your
# app and run "pip install -r requirements.txt" in the Dockerfile instead.
```

PR #106, opened 2019-08-08
Small fixes as I worked through the sample.

SurferJeffAtGoogle:
Added a .csproj file and appsettings.json files.

Changed the code so the user doesn't have to set the PORT environment variable to run the app locally.

Changed helloworld-csharp to HelloGKE to make it a little more idiomatic.


gabrielmollibacache:
Hello Jeffrey,

We ran a test on kubernates and then deleted. Could you please tell us if there any action we must do on our environment?

Please pinpoint me some documents to check if there is any kubernate cluster still running.

Thank you.

Regards,

El 8 ago. 2019 6:52 PM, Jeffrey Rennie <notifications@github.com> escribió:

Merged #106 into master.

PR #107, opened 2019-08-30
Improve Quickstart container images (Node.js, Python, Ruby, C#, Java)

PicardParis:


grayside:
This looks good to me from a container practices standpoint, but I don't have context on the existing Dockerfile decisions so holding off on an official LGTM.

PR #132, opened 2020-07-29
add gke-whereami to samples

theemadnes:
Per internal discussion, adding `gke-whereami` to GKE samples to enable public-facing multicluster demos

### DIFF
--- a/gke-whereami/README.md
+++ b/gke-whereami/README.md
@@ -1,0 +1,85 @@
+# gke-whereami
+
+A simple Kubernetes-oriented app for describing the location of the pod serving a request via its attributes (cluster name, cluster region, pod name, namespace, service account, etc). The response payload includes an emoji that is hashed from the pod name, which makes it a little easier for a human to visually identify the pod you're dealing with.
+
+This was originally written for testing & debugging multi-cluster ingress use cases on GKE (now Ingress for Anthos).
+
+### Setup
+
+Create a GKE cluster 
+
+First define your environment variables (substituting where #needed#):
+
+```
+export PROJECT_ID=#YOUR_PROJECT_ID#
+
+export COMPUTE_REGION=#YOUR_COMPUTE_REGION#
+
+export CLUSTER_NAME=whereami
+```
+
+Now create your resources:
+
+```
+gcloud beta container clusters create $CLUSTER_NAME \
+  --enable-ip-alias \
+  --enable-stackdriver-kubernetes \
+  --region=$COMPUTE_REGION \
+  --num-nodes=1 \
+  --release-channel=regular
+
+gcloud container clusters get-credentials $CLUSTER_NAME --region $COMPUTE_REGION
+```
+
+Deploy the service/pods:
+
+```kubectl apply -k k8s```
+
+*or*
+
+```kustomize build k8s | kubectl apply -f -```
+
+Get the service endpoint:
+```
+WHEREAMI_ENDPOINT=$(kubectl get svc | grep -v EXTERNAL-IP | awk '{ print $4}')
+```
+
+Wrap things up by `curl`ing the `EXTERNAL-IP` of the service. 
+
+```curl $WHEREAMI_ENDPOINT -H "Host: hello"```
+
+Result:
+
+```{"cluster_name":"cluster-1","host_header":"34.66.118.115","id_string":"frontend","node_name":"gke-cluster-1-default-pool-c91b5644-v8kg.c.alexmattson-scratch.internal","pod_ip":"10.4.2.15","pod_name":"whereami-c657c68f5-9jtkw","pod_name_emoji":"👇","pod_namespace":"default","pod_service_account":"whereami-ksa","project_id":"alexmattson-scratch","timestamp":"2020-07-29T19:08:31","zone":"us-central1-c"}```
+
+
+#### using gke-whereami to call downstream services 
+
+`gke-whereami` has an optional flag within its configmap that will cause it to attempt to call another backend service within your GKE cluster (for example, a different, non-public instance of itself). This is helpful for demonstrating a public microservice call to a non-public microservice, and then including the responses of both microservices in the payload delivered back to the user.  
+
+*NOTE:* this backend call assumes the downstream service is returning JSON.
+
+First, build the "backend" instance of `gke-whereami`:
+
+```kustomize build k8s-backend-overlay-example | kubectl apply -f -```
+
+*or*
+
+```kubectl apply -k k8s-backend-overlay-example```
+
+Once that service is up and running, modify `k8s/configmap.yaml`'s `BACKEND_ENABLED` to `"True"`. You will have to redeploy the pods in the whereami service as they will not automatically be recreated when you update the configmap.
+
+The (*slightly* busy-looking) result should look like this:
+
+```{"backend_result":{"cluster_name":"cluster-1","host_header":"whereami-backend","id_string":"backend","node_name":"gke-cluster-1-default-pool-c91b5644-q7w5.c.alexmattson-scratch.internal","pod_ip":"10.4.0.10","pod_name":"whereami-backend-966547575-nk9df","pod_name_emoji":"📔","pod_namespace":"default","pod_service_account":"whereami-ksa-backend","project_id":"alexmattson-scratch","timestamp":"2020-07-29T19:32:15","zone":"us-central1-c"},"cluster_name":"cluster-1","host_header":"34.72.90.134","id_string":"frontend","node_name":"gke-cluster-1-default-pool-c91b5644-v8kg.c.alexmattson-scratch.internal","pod_ip":"10.4.2.16","pod_name":"whereami-c657c68f5-r65s5","pod_name_emoji":"🌋","pod_namespace":"default","pod_service_account":"whereami-ksa","project_id":"alexmattson-scratch","timestamp":"2020-07-29T19:32:15","zone":"us-central1-c"}```
+
+If you wish to call a different backend service, modify `k8s/configmap.yaml`'s `BACKEND_SERVICE` to some other service name. 
+
+
+#### Notes
+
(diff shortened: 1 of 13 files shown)


gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```kustomize build k8s-backend-overlay-example | kubectl apply -f -```
+
+*or*
+
+```kubectl apply -k k8s-backend-overlay-example```
+
+Once that service is up and running, modify `k8s/configmap.yaml`'s `BACKEND_ENABLED` to `"True"`. You will have to redeploy the pods in the whereami service as they will not automatically be recreated when you update the configmap.
+
+The (*slightly* busy-looking) result should look like this:
```
askmeegs, line 72: Add the command you used to get this result.

gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+*NOTE:* this backend call assumes the downstream service is returning JSON.
+
+First, build the "backend" instance of `gke-whereami`:
+
+```kustomize build k8s-backend-overlay-example | kubectl apply -f -```
+
+*or*
+
+```kubectl apply -k k8s-backend-overlay-example```
```
askmeegs, line 68: My GKE cluster - as spec-ed out in your README - is not big enough to fit both whereami and whereami-backend: 

```
NAME                                READY   STATUS    RESTARTS   AGE
whereami-66b5bdd564-487dp           1/1     Running   0          53s
whereami-66b5bdd564-kgbhf           1/1     Running   0          53s
whereami-66b5bdd564-s5k8j           1/1     Running   0          53s
whereami-backend-6746f4b95d-4r5d9   0/1     Pending   0          45s
whereami-backend-6746f4b95d-ngtlm   0/1     Pending   0          45s
whereami-backend-6746f4b95d-qgxbs   0/1     Pending   0          45s


 0/1 nodes are available: 1 Insufficient cpu.
```

gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+Now create your resources:
+
+```
+gcloud beta container clusters create $CLUSTER_NAME \
+  --enable-ip-alias \
+  --enable-stackdriver-kubernetes \
+  --region=$COMPUTE_REGION \
+  --num-nodes=1 \
```
askmeegs, line 28: re: backend overlay example, I recommend the cluster has at least 2 nodes

gke-whereami/README.md
askmeegs: Explain what users should be looking for, in the output.

gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
...
+Result:
+
+```{"cluster_name":"cluster-1","host_header":"34.66.118.115","id_string":"frontend","node_name":"gke-cluster-1-default-pool-c91b5644-v8kg.c.alexmattson-scratch.internal","pod_ip":"10.4.2.15","pod_name":"whereami-c657c68f5-9jtkw","pod_name_emoji":"👇","pod_namespace":"default","pod_service_account":"whereami-ksa","project_id":"alexmattson-scratch","timestamp":"2020-07-29T19:08:31","zone":"us-central1-c"}```
+
+
+#### using gke-whereami to call downstream services 
+
+`gke-whereami` has an optional flag within its configmap that will cause it to attempt to call another backend service within your GKE cluster (for example, a different, non-public instance of itself). This is helpful for demonstrating a public microservice call to a non-public microservice, and then including the responses of both microservices in the payload delivered back to the user.  
+
+*NOTE:* this backend call assumes the downstream service is returning JSON.
```
askmeegs, line 60: I would remove this line, and constrain this example to only work with two instances of `whereami`, both with JSON output

gke-whereami/app.py
```diff
@@ -0,0 +1,150 @@
...
+
+# set up emoji list
+emoji_list = list(emoji.unicode_codes.UNICODE_EMOJI.keys())
+
+
+@app.route('/')
+def home():
+
+    # get project ID
+    try:
```
mark-church, line 28: As we discussed, in each try block, check for null value of `os.getenv` and omit the key from the JSON response if it does not exist. The goal is to be able to do this and have the app be fully functioning for the simplest possible demo:

```
kubectl run --image=gcr.io/alexmattson-scratch/whereami --labels app=whereami --expose --port 8080
```

This should still output information that will always be in the pod like timestamp and hostname. If the user wants to do a more complex demo they can simply add just the environment variables they want to the ConfigMap.

gke-whereami/README.md
theemadnes: it's provisioning a regional cluster, so it should result in a node count of 3.

gke-whereami/README.md
theemadnes: ok, i'll take a look - it should have provisioned a cluster with 3 nodes. investigating...

gke-whereami/README.md
theemadnes: via `--region=$COMPUTE_REGION`

gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
+# gke-whereami
+
+A simple Kubernetes-oriented app for describing the location of the pod serving a request via its attributes (cluster name, cluster region, pod name, namespace, service account, etc). The response payload includes an emoji that is hashed from the pod name, which makes it a little easier for a human to visually identify the pod you're dealing with.
+
+This was originally written for testing & debugging multi-cluster ingress use cases on GKE (now Ingress for Anthos).
+
+### Setup
+
+Create a GKE cluster
```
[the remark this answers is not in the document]
theemadnes: added, but not sure it matches the desired formatting

gke-whereami/README.md
theemadnes: updated per-zone node count to 2

gke-whereami/app.py
theemadnes: done

gke-whereami/app.py
```diff
@@ -0,0 +1,146 @@
...
+if __name__ == '__main__':
+    out_hdlr = logging.StreamHandler(sys.stdout)
+    fmt = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+    out_hdlr.setFormatter(fmt)
+    out_hdlr.setLevel(logging.INFO)
+    logging.getLogger().addHandler(out_hdlr)
+    logging.getLogger().setLevel(logging.INFO)
+    app.logger.handlers = []
+    app.logger.propagate = True
+    app.run(host='0.0.0.0', port=int(os.environ.get('PORT', 8080)))
```
mark-church, line 146: Just an idea ...

Would it be better to run and expose this app on port 80? 

I just tried the app out with the one liner and here is what I did:

```
$ kubectl run --image=gcr.io/alexmattson-scratch/whereami --labels app=whereami --expose --port 8080 foo-whereami

# Tested out curling from another pod inside the cluster
$ kubectl exec -it netshoot-794598cdc4-x67z9 -- sh
/ # curl 10.8.7.222
...
...

# Spent 1 min trying to figure out why app was not working. Then realized that I should have done curl 10.8.7.222:8080
```

If it runs on port 80, is exposed on port 80, then it makes curling it for quick demos even easier and more automatic. Also, using consistent port in the app and also for pod exposure makes it simpler to understand because you're not trying to map different ports in your head. This is a demo-focused app so all of these things help to make the demos as simple and clean as they can be.

mark-church: Also awesome work on making the one-liner work! That is so easy and automatic that it will make this great for tutorials.

theemadnes: chatted about this in person already, but the issue with port 80 is around permissions. unless the app runs as root within the container, you'll get a permission denied error: ```PermissionError: [Errno 13] Permission denied```

gke-whereami/README.md
theemadnes: discussed with @askmeegs and it's a validation issue with `gcloud`

gke-whereami/app.py
```diff
@@ -0,0 +1,150 @@
...
+    # get the whereami ID_STRING envvar
+    id_string = os.getenv('ID_STRING')
+
+    payload = {}
+    payload['cluster_name'] = cluster_name
+    payload['host_header'] = host_header
+    payload['node_name'] = node_name
+    payload['pod_ip'] = pod_ip
+    payload['pod_name'] = pod_name
+    payload['pod_name_emoji'] = emoji_list[hash(pod_name) % len(emoji_list)]
```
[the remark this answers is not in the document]
theemadnes: yes - it's to help visually identify which pod is responding to your request. easier than looking for some string.

gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Deploy the service/pods:
+
+```kubectl apply -k k8s```
+
+*or*
+
+```kustomize build k8s | kubectl apply -f -```
+
+Get the service endpoint:
```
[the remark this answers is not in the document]
theemadnes: done

gke-whereami/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+WHEREAMI_ENDPOINT=$(kubectl get svc | grep -v EXTERNAL-IP | awk '{ print $4}')
```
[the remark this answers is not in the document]
theemadnes: done

gke-whereami/k8s/deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+  template:
+    metadata:
+      labels:
+        app: whereami
+        version: v1
+    spec:
+      serviceAccountName: whereami-ksa
+      containers:
+      - name: whereami
+        image: gcr.io/alexmattson-scratch/whereami@sha256:b592d113dff6ca1584d8f8d5a3aa2aca437e1764b2c86db061d9a7eaa3e81815
```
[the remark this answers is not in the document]
theemadnes: now references `v1.0` instead of a hash

PR #134, opened 2020-08-07
Added header echo feature flag 

theemadnes:
updated README and `app.py` to enable header echo'ing


theemadnes:
@mark-church FYI

PR #136, opened 2020-08-11
Updated pod image references to GoogleCloud & fixed bug in kustomize file for echo-headers

theemadnes:
Tested new image location for all 5(!) deployment modes:

- standalone pod/svc
- kustomize -> k8s/
- kustomize -> frontend overlay
- kustomize -> backend overlay
- kustomize -> echo headers 

Also, there was a bug in the `kustomization` file for `echo-headers` that referenced a different deployment model, so that was patched as well.


theemadnes:
looks like I forgot to pull down the latest PR (which included my own updates). but the conflicts are safe to resolve. CC: @askmeegs @mark-church

PR #155, opened 2020-11-02
Adding Ingress Manifests

melissakendall:
This PR is to create copies of the manifests in the GKE tutorial "Setting up HTTP(S) Load Balancing with Ingress"[1] into GitHub. 

Note that the current tutorial instructs the user to replace the `basic-ingree.yaml` manifest file with the content of a new file. My suggestion is to split that into two files, and change the tutorial to specify to use the new manifest instead of copying it[2]. Alternatively, I could omit importing the file as it currently uses line-highlighting on the annotation that is likely not possible to do when pulling the file from github.

If you agree with this suggestion, I can update the tutorial after this PR is merged.

Prior to merging, I will squash the commits into a single, descriptive commit.

[1] https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer
[2] https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer#step_5_optional_configure_a_static_ip_address

### DIFF
--- a/load-balancing/README.md
+++ b/load-balancing/README.md
@@ -4,5 +4,4 @@ This example shows how to run a web application behind
 an [external HTTP(S) load balancer](https://cloud.google.com/load-balancing/docs/https)
 using the [Ingress](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress) resource.
 
-Visit https://cloud.google.com/kubernetes-engine/docs/concepts/ingress
-to follow the tutorial.
+Visit https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer to follow the tutorial.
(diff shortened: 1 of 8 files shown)


askmeegs:
Hi Melissa, thanks for this! Two separate Ingress files sounds like a good plan.

Would it also be possible to add the Deployments and Services used in the tutorial to this directory? 

https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer/web-deployment.yaml
https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer/web-service.yaml
https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer/web-deployment-v2.yaml
https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer/web-service-v2.yaml

melissakendall:
Thank you @askmeegs I had not noticed those, I will add them

askmeegs (approved):

PR #164, opened 2021-02-08
Update service.yaml

ameer00:
the contenst are indented in the YAML and thus fails when trying to apply service.yaml.


theemadnes (approved):
LGTM - thanks for catching this.

PR #165, opened 2021-02-11
and per-field URLs for payload to whereami

theemadnes:
per request from @mark-church 
```
Random Q ... are there any endpoints for which I can get just a subset of the information that whereami provides
like
curl ${VIP}/pod_name
bar-v2-6bbbd94c98-f4kvv
```

So now if you hit a `whereami` pod at some path, if the end of the path matches a field name in the JSON payload `whereami` typically returns, it gets returned as plaintext. Ex:

```bash
$ curl $ENDPOINT/some/path/prefix/pod_name_emoji

🧚🏽
```

### DIFF
--- a/whereami/README.md
+++ b/whereami/README.md
@@ -193,6 +193,17 @@ $ curl $ENDPOINT
 }
 ```
 
+The JSON payload example above covers the majority of fields that `whereami` can return. In the following sections, you will see how it's possible to have `whereami` to call downstream services, adding additional data to that payload. Before you do that, it's worth pointing out that `whereami` can also return individual fields from that JSON payload as plaintext, so long as you include the field's name as a suffix to the path you're calling. Let's see an example.
+
+Suppose you only care about the `pod_name_emoji` value. You can do the following to capture only that value in the response:
+
+```bash
+$ curl $ENDPOINT/path12345/pod_name_emoji
+
+🧚🏽
+```
+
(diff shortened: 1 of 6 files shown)


whereami/README.md
```diff
@@ -193,6 +193,24 @@ $ curl $ENDPOINT
 }
 ```
 
+The JSON payload example above covers the majority of fields that `whereami` can return. In the following sections, you will see how it's possible to have `whereami` to call downstream services, adding additional data to that payload. Before you do that, it's worth pointing out that `whereami` can also return individual fields from that JSON payload as plaintext, so long as you include the field's name as a suffix to the path you're calling. Let's see an example.
+
+Suppose you only care about the `pod_name_emoji` value. You can do the following to capture only that value in the response:
+
+```bash
+$ curl $ENDPOINT/path12345/pod_name_emoji
```
askmeegs, line 201: is `path12345` intended here? I tried testing this with just the field name in the path, and worked for me -- 

```
❯ curl $ENDPOINT/pod_name_emoji
🧑🏿‍🎤%
```

theemadnes: it is - yeah it was just meant to reflect some arbitrary prefix in the path. should i change it to `some/path` or something more generic like that?

theemadnes: ok updated to this instead:

```bash
$ curl $ENDPOINT/some/path/prefix/pod_name_emoji

🧚🏽
```

askmeegs (approved):

PR #183, opened 2021-05-13
Bump gke-windows-builder version to 2.5.0

pjh:
Fixes from version 2.0.0 -> 2.5.0 include:

- Add Windows Server 2004 and 20H2 arch's, remove 1909 arch.
- Add boot-disk-type flag.
- Add logic to create a bucket to copy workspace if not exists
- Move the Winrm setup to last step to reduce the chance of failures
- Add bootDiskSizeGB option
- Disable Windows Defender to avoid accasional docker build failures
- Fail the building process if errors happen


pjh:
cc @jeremyje

jeremyje (approved):

PR #190, opened 2021-06-17
Revert "fix: update dockerfile to use composer instead of pear (#171)"

jimangel:
This reverts commit 41c22011b7b66672f9815bd762fbbc4d7e96ef72.

There's a long story of how I got here, but the TL;DR is: #171 is a breaking change that results in https://cloud.google.com/kubernetes-engine/docs/tutorials/guestbook and other tutorials not working due to the missing `Predis` package.

I don't know enough about the code to fix using pear or alternative solution, but a revert does fix it and I've proved it out w/ [this personal publicly built image](https://hub.docker.com/repository/docker/jimangel/php-guestbook-revert).

Without the revert, php never finds `Predis` and errors out (no data / connection):

image

With the revert, php finds `Redis` and things work:

image

I was able to connect the dots when the error I googled led me to a [discussion around `compose install` vs `compose upgrade`](https://stackoverflow.com/a/41209642) and [the PR](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/171#issue-585048759) had comments around the same.

There's probably a better solution using `compose`, but I don't have the time to figure it out and this is a way to get things operational again. Thanks!

/cc @gabidavila

### DIFF
--- a/guestbook/php-redis/Dockerfile
+++ b/guestbook/php-redis/Dockerfile
@@ -11,10 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM php:7.4-apache-buster
-RUN apt update && apt -y install zip unzip git-all
-COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
+FROM php:7.4.7-apache
 
-COPY . .
+RUN pear channel-discover pear.nrk.io
+RUN pear install nrk/Predis
 
-RUN composer update
+ADD guestbook.php /var/www/html/guestbook.php
+ADD controllers.js /var/www/html/controllers.js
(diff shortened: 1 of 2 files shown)


jimangel:
@askmeegs / @donmccasland any concerns about merging this PR?

I believe, at this the time, both https://kubernetes.io/docs/tutorials/stateless-application/guestbook/ and https://cloud.google.com/kubernetes-engine/docs/tutorials/guestbook are broken demos due to the missing Predis package.

NimJay:
Hi @jimangel!

First off, thanks so much for investigating and making this pull-request! :)
But I didn't notice this issue (and your PR) until yesterday. My apologies.

The error was a result of us incorrectly importing the "Predis" PHP library.
We have fixed it in https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/195. 
Hope this works for you.

Closing this pull-request now.

PR #193, opened 2021-06-22
Add Terraform for Google Cloud Build Triggers

NimJay:
This pull-request:

1. adds a `terraform/` folder.
1. adds a `google-cloud-build-trggers.tf` file which contains the Terraform configurations for the Google Cloud Build Triggers that build/push Docker images to Container Registry and Artifact Registry. 
1. adds a `.gitignore` file to prevent future developers from committing Terraform-related files that [shouldn't be committed](https://stackoverflow.com/a/38748987).
1. ~adds the `.terraform.lock.hcl` that was generated by Terraform (to ensure we all use the same version of the Google Terraform plugin). [Terraform recommends](https://www.terraform.io/upgrade-guides/0-14.html#opting-out-of-dependency-locking) committing this file.~ The `.terraform.lock.hcl` file is no longer included — as suggested below by @donmccasland.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+# Terraform
+**/.terraform/*
+*.tfstate
+*.tfstate.*
--- a/terraform/google-cloud-build-triggers.tf
+++ b/terraform/google-cloud-build-triggers.tf
@@ -5,8 +5,16 @@
 // How to use this file:
 //     1. Install Terraform.
 //     2. From this directory, run "terraform init". This will download the Google Terraform plugin.
+//        If you get an error similar to "querying Cloud Storage failed: storage: bucket doesn't exist",
+//        try running: gcloud auth application-default login
 //     3. Finally, run "terraform apply" to create any missing Google Cloud Build Triggers.
-//        You'll see errors for the Google Cloud Build Triggers that already exist.
+
+terraform {
+  backend "gcs" {
+    bucket = "kubernetes-engine-samples"
+    prefix = "terraform-state"
+  }
+}
 
 provider "google" {
     project = "google-samples"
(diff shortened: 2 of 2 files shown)


terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
...
+    "zh:8b15d9c37cd04303f866b66bffe13c15d8e3254459958ebaad184d63dc1f5da0",
+    "zh:920b8b65a6900057c664cb099d64cc58f88ca4ab38e13e2c805cf2413a82795b",
+    "zh:bd022571709bff30d9d18a1bb2a383df85027b01ffbbf92ce9e514d48557899e",
+    "zh:c68b306b61e637a66f8c20580f23143517f957ed74b52476e4094a14fd72e615",
+    "zh:e5f2fe60bc99612cc089265d7003f6d2007304ec37367a43eb65716de2d1f491",
+    "zh:e865d87acf9c357b12ed55538d8d974a708762ed6288119e4c33591cae70159e",
+    "zh:f7b52e22b86f11a06f822a9b1176a4d50e120a6b4cfce61a0574007b5d143370",
+    "zh:ff98905d147426cfc5b018184706774e323993f773333802999eec9974c58773",
+  ]
+}
```
donmccasland, line 21: I think that typically folks do not check in the .terraform.lock.hcl, as it is generated by terraform init.  While this may allow us to lock the version of terraform, it would also keep us from knowing how latest terraform versions perform against our code.

donmccasland (approved):

donmccasland (commented):
We could also create a backend.tf that would store tf state in a GCS bucket like this:
terraform {
  backend "gcs" {
    bucket  = "tf-state"
    prefix  = "terraform/state"
  }
}

NimJay:
We are now storing the Terraform state in Google Cloud Storage (GCS)!
* I created a GCS bucket called `kubernetes-engine-samples` in the `google-samples` project.
* The Terraform state is now stored in that bucket (see most recent commit: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/193/commits/35b4e0f37d6ac8e691f09bd2ef7476aab6dc9232). 
* I have deleted the old manually-created Triggers (in the `google-samples` project). The Triggers (see [Triggers here](https://pantheon.corp.google.com/cloud-build/triggers?project=google-samples)) are now created as a result of the Terraform config from the pull-request.

donmccasland (approved):

Issue #170, opened 2021-03-04
XSS Vulnerability on the Guestbook example for php-redis


gabidavila:
Someone reached out to me about this possible path of exploitation through XSS on this part of the code:

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/guestbook/php-redis/guestbook.php#L39

`$_GET` here is not recommend, since you could override any value just by guessing a key, or setting arbitrary keys.

The recommendation I make it is maybe to do an API endpoint?

NimJay:
Hi @gabidavila.

First, thanks for bringing up this concern!

I realize that it's been a long time since you reported this issue — so I'm sorry to make you jog your memory back so far.
Here's a refresher:
* This sample app just keeps track of a list of names (a guestbook).
* It comprises of: 
  * an AngularJS webpage; 
  * a PHP `guestbook.php` HTTP endpoint; and 
  * a Redis key-value database.
* The guestbook names are stored in the Redis key-value database, under the `messages` key like so: `"Katy Perry, John Lenon, Paul McCartney"`.
* The a PHP `guestbook.php` HTTP endpoint allows for retrieval and modification of values at _any_ key — not just the `messages` key.
* The AngularJS webpage talks to the `guestbook.php` endpoint to save and retrieve names (the guestbook). 

If possible, could you please elaborate on what you mean by:
> The recommendation I make it is maybe to do an API endpoint?
Do you mean we should replace `guestbook.php` with an HTTP endpoint that only allows for retrieval and modification of values at the `messages` key?

Let me know if you have any questions (and feel free to ping me on Google Chat).
Thank you!

NimJay:
I'm closing this issue as it has been addressed by https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/202.
@gabidavila Feel free to ignore the above question(s). Thank you! :)

PR #202, opened 2021-08-10
Remove guestbook.php's redundant 'key' GET parameter

NimJay:
This addresses https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/170.

**Background**
* The Guestbook sample app consists of:
  1. Redis key-value database - stores the list of names (i.e., guestbook).
  1. `guestbook.php` HTTP endpoint - allows for retrieval and modification of values at _any_ Redis key.
  1. AngularJS webpage - makes HTTP GET requests to `guestbook.php` (to store a string like `Jill Doe, John Doe, Jane Doe` in the Redis key-value store at key `messages`).

**Change Summary**
* `guestbook.php` now only allows for retrieval and modification of values at key `guestbook` — instead of allowing any key.
* The AngularJS webpage no longer specifies (to `guestbook.php`) the Redis key (`messages`) in which the guestbook (string of names) is stored.

**Things to Consider**
* Merging this pull-request will invoke a Cloud Build Trigger that rebuilds/repushes the Docker image at `gcr.io/google_samples/gb-frontend:v5`.  Perhaps we should increment `v5` to `v6` before merging.
  * In practice, one would increment `v5` to `v6`. But I don't think we need to in this case, because this doesn't change how our users interface with the tutorial.

**Manual Testing Passed**
* I have manually tested the changes in my own GKE cluster, and it works. 
Screen Shot 2021-08-10 at 1 37 34 PM

### DIFF
--- a/guestbook/php-redis/controllers.js
+++ b/guestbook/php-redis/controllers.js
@@ -25,7 +25,7 @@ RedisController.prototype.onRedis = function() {
     this.scope_.messages.push(this.scope_.msg);
     this.scope_.msg = "";
     var value = this.scope_.messages.join();
-    this.http_.get("guestbook.php?cmd=set&key=messages&value=" + value)
+    this.http_.get("guestbook.php?cmd=set&value=" + value)
             .success(angular.bind(this, function(data) {
                 this.scope_.redisResponse = "Updated.";
             }));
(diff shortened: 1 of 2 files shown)


bourgeoisor (approved):
LGTM.
We could also remove the `cmd` param by utilizing both `GET` and something like `PUT`, but that's out of scope of this bug fix.
As for the image tag, I think it's fine to leave it as-is since, as you mentioned, these are minor changes with nothing breaking from a user perspective.

NimJay:
I have tested the new Docker image created by [the Google Build Trigger](https://pantheon.corp.google.com/cloud-build/builds/6c59e017-6a52-48b9-810b-b9208f77d360;step=0?project=google-samples) (upon merging).
It works! ✅ 
I basically just tested the relevant tutorial, [Create a guestbook with Redis and PHP](https://cloud.google.com/kubernetes-engine/docs/tutorials/guestbook).

Screen Shot 2021-08-11 at 10 16 21 AM

PR #206, opened 2021-09-02
Add workload metrics sample

mastersingh24:
This sample leverages the GKE workload metrics pipeline to scale a sample application using an HPA.
It is the reference  source for an upcoming GKE tutorial.


workload-metrics/manifests/workload-metrics-deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+  selector:
+    matchLabels:
+      app: workload-metrics-example
+  template:
+    metadata:
+      labels:
+        app: workload-metrics-example
+    spec:
+      containers:
+      - image: nilebox/prometheus-example-app@sha256:dab60d038c5d6915af5bcbe5f0279a22b95a8c8be254153e22d7cd81b21b84c5
```
askmeegs, line 18: where does the code for this live? is there an image tag you could use here instead of a SHA?

workload-metrics/README.md
```diff
@@ -0,0 +1,9 @@
+# Workload Metrics example
+
+This tutorial demonstrates how to automatically scale your Google Kubernetes Engine (GKE)
```
askmeegs, line 3: Hello Gari, I'm a bit confused about how this is different from the existing Custom Metrics Autoscaling tutorial and sample. 

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/tree/master/custom-metrics-autoscaling 

https://cloud.google.com/kubernetes-engine/docs/tutorials/autoscaling-metrics

workload-metrics/manifests/workload-metrics-hpa.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: autoscaling/v2beta2
```
askmeegs, line 1: All YAML samples need licenses and region tags. 
go/code-snippets-101 

Example: 
https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/hello-app/manifests/helloweb-deployment.yaml

workload-metrics/manifests/workload-metrics-podmon.yaml
```diff
@@ -0,0 +1,31 @@
+# Note that this PodMonitor is in the monitoring.gke.io domain,
+# rather than the monitoring.coreos.com domain used with the
+# Prometheus Operator
+apiVersion: monitoring.gke.io/v1alpha1
+kind: PodMonitor
```
askmeegs, line 5: Where is the documentation for this resource? 
Is this feature launched?

askmeegs (changes requested):
A few questions.

workload-metrics/manifests/workload-metrics-deployment.yaml
mastersingh24: I actually moved all of the code into this repo now and included a cloudbuild.yaml to tag a 1.0 image and push to google-samples

workload-metrics/README.md
mastersingh24: I updated the text here as well, but basically, GKE has a new feature called "GKE workload metrics" which allows you to configure a policy to scrape Prometheus metrics from pods and send them to Cloud Monitoring.  So they are no longer "custom metrics" per se but rather "workload metrics" with their own Cloud Monitoring prefix.

We still need to deploy the adapter for HPA to work, but no longer need the prom-sd sidecar or to use sd metrics in the app.

workload-metrics/manifests/workload-metrics-podmon.yaml
mastersingh24: We have a chicken and egg problem ... the code is available but the blog and docs have not yet been released.

askmeegs (approved):

mastersingh24:
Looks great!  Thanks for the assistance!

Issue #207, opened 2021-09-07
Update whereami utility to be cloud-agnostic


theemadnes:
@knee-berts brought up an excellent point - with Anthos being multi-cloud, `whereami` should be able to return the same payload regardless of which supported platform it's running on. Currently investigating implementation approaches with him.

theemadnes:
potential platforms include GCP, AWS, Azure, etc

theemadnes:
currently `whereami` references the GCE metadata endpoint to pull some info, so effectively we need to come up with a non-GCP alternative

theemadnes:
Also potentially in scope: non-cloud platforms like bare metal & vSphere.

askmeegs:
Hi Alex, I agree this is a good idea that could help with Anthos multicloud demos. 
What happens today when this app runs on AWS or Azure? Does it throw an error trying to call the GCE metadata endpoint? 

Because there are two options -- 
1) Fail gracefully if the app can't reach the GCE metadata server. That's what we're doing with the bigger sample apps today. The region just shows up as "unknown" in the frontend. https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/frontend.py#L543 

2) Add AWS / Azure specific code to detect what cloud provider the user is running in, and call the appropriate server to fetch the region/zone.  But this is a bit of a rabbit hole - ie. will you also cover baremetal, kind, and all the other Anthos supported environments?  Arguably then, this app would not belong in this repo (GKE on GCP samples) but rather  https://github.com/GoogleCloudPlatform/anthos-samples which covers the non GCP environments. 

It may be also worth asking the Anthos multicloud team if they're using any in-house apps to demo the product / see if something similar has ever been done.

theemadnes:
incremental updated, but the PR i just put in will move the node name source from GCE metadata to Downward API, which makes it work on any K8s environment

askmeegs:
Thanks Alex! We might try this workaround in the Bank of Anthos and Online Boutique sample apps.

theemadnes:
cool. happy to contribute those changes if you're looking to offload some work. LMK.

theemadnes:
also @askmeegs the latest PR doesn't totally solve the incomplete payload issue - it just makes the node_name field work, which hopefully is sufficient for most use cases. the fields that won't work outside of GCP after this PR are as follows:

- `project_id` of course
[...]

PR #211, opened 2021-09-23
NodeName from Downward API

theemadnes:
related to the discussion in https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/207 moving the node name data source to downward API instead of pulling from GCE metadata server.

### DIFF
--- a/whereami/README.md
+++ b/whereami/README.md
@@ -11,7 +11,7 @@
 `whereami` is a single-container app, designed and packaged to run on Kubernetes. In its simplest form it can be deployed in a single line with only a few parameters.
 
 ```bash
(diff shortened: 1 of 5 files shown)


askmeegs:
Looks good! One request, can you update the CI to build v1.2.2 instead of v1.2.1 ? This way you shouldn't have to manually build + push the images anymore -we have a cloud build trigger to do that. thanks! 

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/master/whereami/cloudbuild.yaml#L24

theemadnes:
done

askmeegs (approved):

Issue #224, opened 2021-11-10
pub-sub Docker image Python 3.10 breaking change


mihaidraghici98:
Hello!

There's an issue with `gcr.io/google-samples/pubsub-sample:v1` image.

It fails on startup due to MutableMaping not being found in collections library.
In Python3.10 there is a breaking change which removed collections.MutableMaping alias which poited to collections.abc.MutableMaping:

> Aliases to Abstract Base Classes in the collections module, like collections.Mapping alias to collections.abc.Mapping, are kept for one last release for backward compatibility. They will be removed from Python 3.10.
[What's new in Python3.9](https://docs.python.org/3/whatsnew/3.9.html)

A quick solution that I have found is changing the base image in the Dockerfile [here](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/718e09b69a1224e17930354d6e0eff4e18a5e67f/cloud-pubsub/Dockerfile#L1) to python:3.8-slim

NimJay:
Hello @mihaidraghici98.
Thank you so much for reporting this issue and suggesting a solution. :)

@ckim328  
So sorry! 😅 
I misread and misunderstood the issue — I thought `docker build`-ing of the `pubsub-sample:v1` was broken. But now I understand that deployment of the image is broken. Your initial "p1" label was definitely correct. I was wrong when I suggested that this is a "p2".

I am currently reviewing #225.

NimJay:
This issue is now fixed. See #225.
Thanks again, both of you! :)

PR #225, opened 2021-11-11
Specified python3.8-slim image in Dockerfile

ckim328:
This addresses [this issue](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/224), where  a breaking change was introduced in Python 3.10.

### DIFF
--- a/cloud-pubsub/Dockerfile
+++ b/cloud-pubsub/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3-slim
+FROM python:3.8-slim
 COPY . .
 RUN pip install -r requirements.txt
 ENTRYPOINT ["python", "-u", "main.py"]


NimJay:
Thanks for the quick this pull-request, @ckim328! (And sorry for the late review.)

Judging from [What’s New In Python 3.9](https://docs.python.org/3/whatsnew/3.9.html#you-should-check-for-deprecationwarning-in-your-code), it sounds like we could even use Python 3.9.
> Python 3.9 is the last version providing those Python 2 backward compatibility layers... Aliases to Abstract Base Classes in the collections module, like collections.Mapping

But I'm happy with using the `python:3.8-slim` base image, given how small this sample app is and given the urgency of this issue. 

## Manually Tested 👍 
* I tested the Docker image via the tutorial, [Authenticating to Google Cloud with service accounts](https://cloud.google.com/kubernetes-engine/docs/tutorials/authenticating-to-cloud-platform#kubectl).
* I built and pushed 2 images to my own Google Container Registry:
  1. `pubsub-sample:v1-broken` - This used base image `python:3-slim`. I was able to reproduce the issue and see the `AttributeError: ... 'MutableMapping'` error.
  1. `pubsub-sample:v1-fixed` - This used base image `python:3.8-slim`. The issue was fixed!

NimJay (approved):
Approved! See my previous comment — my testing confirmed that this fix works.

NimJay:
## After Merging

Merging of this pull-request triggers the rebuilding on the Google Cloud Build [Trigger for the `pubsub-sample:v1` image](https://pantheon.corp.google.com/cloud-build/builds/1bce868a-5af1-4a45-94fb-8643a41d7e43?project=google-samples).

The following 2 images have been rebuilt and repushed: 
* `us-docker.pkg.dev/google-samples/containers/gke/pubsub-sample:v1`
* `gcr.io/google-samples/pubsub-sample:v1`

### Manually Tested 👍 

* I did a very quick test of the images at both URLs.
* I just `docker pull`-ed the images and used `docker run`.
* I did not get the `AttributeError: ... 'MutableMapping'`.
* Instead, I got this error which **is expected**:
```
raise exceptions.DefaultCredentialsError(_HELP_MESSAGE)
google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://cloud.google.com/docs/authentication/getting-started
```

Issue #209, opened 2021-09-10
Use AR instead of GCR for Deployments


NimJay:
We are trying to encourage customers to use AR (Artifact Registry) instead of GCR (Google Container Registry).
So all [Deployments](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) in our GKE samples (in this repo) need to use AR-hosted images instead of GCR-hosted images.

## Example
For instance, the following `hello-app:1.0` Deployment 
https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/75e3523f6f0520864524c54cb7cb89c09932eb86/hello-app/manifests/helloweb-deployment.yaml#L35 should change to:
```yaml
image: us-docker.pkg.dev/google-samples/containers/gke/hello-app:1.0
```

## TODOs
Here are the files that need to be updated (and their associated pull-requests):

- [x] [migrating-node-pool/node-pools-deployment.yaml](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/75e3523f6f0520864524c54cb7cb89c09932eb86/migrating-node-pool/node-pools-deployment.yaml)
  * #210
- [x] [cloud-pubsub/deployment/pubsub-with-secret.yaml](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/75e3523f6f0520864524c54cb7cb89c09932eb86/cloud-pubsub/deployment/pubsub-with-secret.yaml)
  * #220
- [x] [cloud-pubsub/deployment/pubsub.yaml](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/75e3523f6f0520864524c54cb7cb89c09932eb86/cloud-pubsub/deployment/pubsub.yaml)
  * #220
- [x] [custom-metrics-autoscaling/direct-to-sd/custom-metrics-sd.yaml](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/75e3523f6f0520864524c54cb7cb89c09932eb86/custom-metrics-autoscaling/direct-to-sd/custom-metrics-sd.yaml)
  * #226 
[...]

PR #259, opened 2022-01-01
Use AR instead of GCR for whereami

irvifa:
* Please see [Use AR instead of GCR for Deployments #209](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/209).

    * AR = Artifact Registry

    * GCR = Google Container Registry

    * This pull-request for `whereami:v1.2.6`:

    * I'm replacing instances of the images' GCR URLs with the equivalent AR URLs.

      * `gcr.io/google-samples/whereami:v1.2.6` → `us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.6`

So once merged, I would have to update the above tutorials via [this Google-internal content publisher](https://devsite.corp.google.com/publisher).

We can verify that the AR (replacement) images and GCR (replaced) images are equivalent like so:

    1. Pull the 2 images you want to compare:

```
docker image pull gcr.io/google-samples/whereami:v1.2.6
docker image pull us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.6
```

    2. Inspecting both images:

```
docker image inspect gcr.io/google-samples/whereami:v1.2.6
docker image inspect us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.6
```

    3. See that the [RootFS values match](https://stackoverflow.com/questions/46321878/how-to-verify-if-the-content-of-two-docker-images-is-exactly-the-same/46322160#46322160).

### DIFF
--- a/whereami/k8s-grpc/deployment.yaml
+++ b/whereami/k8s-grpc/deployment.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: whereami-grpc-ksa
       containers:
       - name: whereami-grpc
-        image: gcr.io/google-samples/whereami:v1.2.6
+        image: us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.6
         ports:
           - name: grpc
             containerPort: 9090
--- a/whereami/k8s/deployment.yaml
+++ b/whereami/k8s/deployment.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: whereami-ksa
       containers:
       - name: whereami
-        image: gcr.io/google-samples/whereami:v1.2.6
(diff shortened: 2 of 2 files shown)


irvifa:
Related to https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/209

NimJay (approved):
Thank you so much for preparing this pull-request, @irvifa! 👏 

The `RootFS` values of the images match.
We will also need to update [`whereami/README.md`](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/tree/main/whereami#simple-deployment); that is, we'll need to replace the instances of `gcr.io/google-samples/whereami:v1.2.6`. I will prepare a separate pull-request for this. 

Approved. And merging.

PR #266, opened 2022-01-19
Adding the grpc-hc-proxy to the GKE samples.

boredabdel:
This PR adds the gRPC HealchCheck proxy forked from salrashid123@ repo. The reason i'm adding this here is to be able to push the images (the proxy and a sample gRPC Server) to the google-samples Artifact Registry. I will maintain the fork going forward.

This proxy is needed for our gke-network-recipes repo (https://github.com/GoogleCloudPlatform/gke-networking-recipes)


askmeegs:
Closing this PR (see internal discussion) as its size + scope warrants a separate repo, and the sample does not match the other samples in this repository, which are intended for embedding in cloud.google.com documentation.

PR #271, opened 2022-01-31
Update Windows cloudbuild config

pjh:
This sample needs to pull the builder from a different artifact registry location. 

Also, follows up on https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/222: we do not want to hard-code the region name or location in this file, the user should fill them in.

This PR aligns with https://cloud.google.com/kubernetes-engine/docs/tutorials/building-windows-multi-arch-images (and the changes I'm making to it in internal change 425411666).

PR #273, opened 2022-02-01
Fix GKE Windows CODEOWNERS again

pjh:
For some reason on https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/271 GitHub is only detecting mcshooter as a code owner, and not ibabou. The owners detection on files for this subdirectory seems to be incomplete:
image

Michelle is on leave and can't approve that PR, so I'm trying to fix the CODEOWNERS by removing mcshooter and liurupeng. liurupeng is on the team but is not a GoogleCloudPlatform org member yet, so perhaps that's tripping up the GitHub owners detection somehow. I don't see any other good reason why ibabou would not have owner approval power with the current file.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,4 +6,4 @@
 *                   @GoogleCloudPlatform/anthos-dpe
 
 # Directory-specific owners
-/windows-multi-arch/ @ibabou @liurupeng @mcshooter @pjh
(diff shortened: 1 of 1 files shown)


ckim328:
Hi @pjh, I updated the settings in the collaborators settings for @ibabou, he should have access now!
Screen Shot 2022-02-01 at 1 28 11 PM

Issue #281, opened 2022-03-16
`Whereami` not starting if GKE node pool's SA doesn't have the Cloud Trace agent role (even with `TRACE_SAMPLING_RATIO` set to `0`)


mathieu-benoit:
When deploying `Whereami`, it's not starting if my GKE node pool's SA doesn't have the Cloud Trace agent role (even with `TRACE_SAMPLING_RATIO` set to `0`).

I got this error in the `Deployment`/`Pod` logs:
```
Traceback (most recent call last): File "/app/app.py", line 51, in <module> cloud_trace_exporter = CloudTraceSpanExporter() File "/usr/local/lib/python3.9/site-packages/opentelemetry/exporter/cloud_trace/__init__.py", line 113, in __init__ self.client: TraceServiceClient = client or TraceServiceClient() File "/usr/local/lib/python3.9/site-packages/google/cloud/trace_v2/services/trace_service/client.py", line 398, in __init__ self._transport = Transport( File "/usr/local/lib/python3.9/site-packages/google/cloud/trace_v2/services/trace_service/transports/grpc.py", line 155, in __init__ super().__init__( File "/usr/local/lib/python3.9/site-packages/google/cloud/trace_v2/services/trace_service/transports/base.py", line 108, in __init__ credentials, _ = google.auth.default( File "/usr/local/lib/python3.9/site-packages/google/auth/_default.py", line 575, in default raise exceptions.DefaultCredentialsError(_HELP_MESSAGE) google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://cloud.google.com/docs/authentication/getting-started
```

I got 2 workarounds:
- Either grant the Cloud Trace agent role to my GKE's SA
- Or create a dedicated GSA to my `Whereami` app and bind it to associated KSA and grand it the Cloud Trace agent role

CC: @theemadnes 

FYI, here is how I create my cluster:
```
gkeSaName=$clusterName-sa
gkeSaId=$gkeSaName@$projectId.iam.gserviceaccount.com
gcloud iam service-accounts create $gkeSaName \
  --display-name=$gkeSaName
roles="roles/logging.logWriter roles/monitoring.metricWriter roles/monitoring.viewer"
[...]

PR #297, opened 2022-04-03
Whereami fix tracing perms

theemadnes:
A few fixes for `whereami` to address https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/281 

- set default to no tracing to Cloud Trace in configmap
- added logic to not load tracing when trace ratio (the default) set to 0
- optimized call to metadata server by switching from 3 calls to a single call and then just handling JSON locally


ckim328:
Thanks so much @theemadnes for getting this ready!

ckim328 (approved):

PR #327, opened 2022-07-12
Fix redis image version and port name

EvanSimpson:
The image version was placed in the port name field rather than the container image field. This moves it to the correct location. Verified locally.


ckim328 (approved):
LGTM! Thanks for catching this 👍

PR #336, opened 2022-09-08
Stateful node pool upgrade for "Upgrading a GKE cluster running a stateful workload"

charlieyu1996:
Changes to the tutorial https://cloud.google.com/kubernetes-engine/docs/tutorials/upgrading-stateful-workload to include more best practices on upgrading a stateful node pool.


bourgeoisor (approved):
LGTM!

PR #340, opened 2022-09-16
add default resources to remove warning in setup

LukeSchlangen:
A warning appears with requests and limits are not set properly (Autopilot still handles everything just fine, but the warning could throw someone off who is following the tutorial). This PR sets the tutorial to match the defaults listed here: https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#defaults

Which removes the warning.


LukeSchlangen:
In draft for now as I go through the [updated Quickstart](https://cloud.google.com/kubernetes-engine/docs/deploy-app-cluster) for Autopilot (checking to make sure this does clear the warning).

quickstart/deployment.yaml
```diff
@@ -39,5 +39,10 @@ spec:
         env:
           - name: PORT
             value: "8080"
+        # Set resources to be the same as GKE Autopilot defaults http://g.co/gke/autopilot-defaults
```
bourgeoisor, line 42: While those are default resources if none specified, they aren't minimum threshold and are probably more than overkill for this workload. Could we do e.g. 250mb of memory and 250m of CPU?

LukeSchlangen:
It seems like this warning might be shown inaccurately (requires setting limits in order to not get the error). I'm going to close this for now since the service should change to remove the warning (shouldn't need to change the tutorial at this point).

LukeSchlangen:
Reopening now that limit bug has been addressed.

bourgeoisor:
This is blocked on an internal bug: 247098340

bourgeoisor:
@LukeSchlangen out of curiosity, do you still get the warning if you set _both_ `requests` and `limits` and that they match to the same values?

quickstart/deployment.yaml
```diff
@@ -39,5 +39,14 @@ spec:
         env:
           - name: PORT
             value: "8080"
+        resources:
+          requests:
+            memory: "2Gi"
```
bourgeoisor, line 44: This seems high-- Is 2Gi necessary for hello-app?

bourgeoisor (approved):

PR #343, opened 2022-09-21
update dotnet example to version 6

LukeSchlangen:
The existing GKE dotnet Quickstart example is from dotnet 2.2 (will fail in cloud shell), this PR will update the example to dotnet 6 (will work in cloud shell).

> Related, but not critical: This example has a reference to Cloud Run for getting the port, but I don't know if that's necessary for GKE? We may want to simplify the code to always use `8080`?

### DIFF
--- a/quickstart/dotnet/Dockerfile
+++ b/quickstart/dotnet/Dockerfile
@@ -1,7 +1,5 @@
 # Use Microsoft's official lightweight .NET images.
-# https://hub.docker.com/r/microsoft/dotnet
-
-FROM mcr.microsoft.com/dotnet/core/sdk:2.2-alpine AS build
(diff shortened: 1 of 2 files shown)


bourgeoisor:
I'm always in favour of simplification; this PR looks great! Has this been tested in the context of the relevant tutorial(s) as well?

LukeSchlangen:
Yes! I walked through the tutorial with these changes. The hardest thing to catch was this line `ENTRYPOINT ["dotnet", "helloworld-gke.dll"]` since the entry point is now the name of the `.csproj` and not the `namespace`.

bourgeoisor (approved):
LGTM

PR #403, opened 2022-11-29
Update Renovate to update Dockerfiles

mathieu-benoit:
Update Renovate to update Dockerfiles, revert back https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/257, needs to be validated by @xtineskim. Could we now update Dockerfiles?

For example, we recently added the build/test in CI (https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/370 + https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/367), is it helping to avoid any issues you faced in the past?

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,7 +3,7 @@
     "config:base"
   ],
   "dockerfile": {
-    "enabled": false
+    "enabled": true
   },
   "pip-compile": {
     "enabled": true,


mathieu-benoit:
Actually closing this one as it will change and impact how most of the apps are built today: once merged in `main` branch. So if Renovates opens a PR, as soon as we will merge this, it will trigger a new build and container image, overriding the current container images version, which are supposed to be immutable... I just checked with Whereami as an illustration of this. We want to keep the manual version upgrades until this release process is in place.

PR #404, opened 2022-11-29
`whereami` - 1.2.14

mathieu-benoit:
In progress...

- Whereami 1.2.14
- Pin Python packages versions as best practices to catch new versions as soon as they come (managed by Renovate)
-


mathieu-benoit:
Actually closing this one because we want to keep the upgrade of Whereami manual to increase the Ma.Mi.Re version. If we pin the Python packages, as soon as a Renovate PR is merged in `main` branch the current Whereami version will be overriden, breaking the immutable concept of that container image...

Issue #426, opened 2022-12-15
improve whereami output when using with Cloud Run / Cloud Functions


theemadnes:
include details like container instance ID and GSA used when running on serverless platform(s) on Google Cloud. 

Pull from: 
https://cloud.google.com/run/docs/container-contract#metadata-server

theemadnes:
Less than ideal but there's a bit of an impedance mismatch when fetching Cloud Run fields since some of the default fields reference "pods" which aren't a thing in CR. ```{
  "cloud_run_instance_id": "0071bb481567295e2eaa3092ab04038a51f7548492e17aa98788ca070921232c81b16bfbad2e14bc928527c8182bb211057c4f2988d01df42206a91b315260e0c9",
  "cloud_run_service_account": "841101411908-compute@developer.gserviceaccount.com",
  "host_header": "whereami-4uotx33u2a-uc.a.run.app",
  "pod_name": "localhost",
  "pod_name_emoji": "8⃣",
  "project_id": "am-arg-01",
  "timestamp": "2022-12-17T02:24:22",
  "zone": "us-central1-1"
}```

theemadnes:
Also the `zone` field isn't entirely accurate as CR is regional in nature; the `-1` suffix is just a default the GCE metadata service applies.

PR #430, opened 2022-12-17
Cloud run fields for whereami

theemadnes:
https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/426

Includes container ID and Cloud Run revision's service account.


whereami/k8s-grpc/deployment.yaml
```diff
@@ -46,7 +46,7 @@ spec:
               - all
           privileged: false
           readOnlyRootFilesystem: true
-        image: us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.14
+        image: us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.15
         ports:
           - name: grpc
             containerPort: 9090
```
mathieu-benoit, line 52: Right below, do you need to change `8080` by `8000` according to your changes in this PR?

mathieu-benoit (changes requested):
Added a comment about port `8000`.

Also, while we are here with a new 1.2.15, could you update the base image in the `Dockerfile` to `python:3.10.9-slim` please?

whereami/k8s/deployment.yaml
```diff
@@ -46,7 +46,7 @@ spec:
               - all
           privileged: false
           readOnlyRootFilesystem: true
-        image: us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.14
+        image: us-docker.pkg.dev/google-samples/containers/gke/whereami:v1.2.15
```
mathieu-benoit, line 49: In this file should you add the `8000` port for prometheus /metrics too?

whereami/whereami_payload.py
```diff
@@ -88,8 +88,15 @@ def call_http_backend(backend_service):
         def call_grpc_backend(backend_service):
 
             try:
-                channel = grpc.insecure_channel(backend_service +
-                                                ':9090')
+                # check for using `443` port to use `secure_channel`
+                # assumes port number is appended to backend_service name
+                if backend_service.split(':')[1] == '443':
```
mathieu-benoit, line 93: `443` is hard coded, could it be `8443` (less privilege) or another port too?

theemadnes: i'll add `8443` and will handle the prometheus port stuff

whereami/k8s-grpc/deployment.yaml
theemadnes: fixing

whereami/k8s/deployment.yaml
theemadnes: yes. can't use `8080` in gRPC mode when using Cloud Run because `8080` is used for the serving port.

mathieu-benoit (approved):

PR #449, opened 2023-01-12
Fix whereami cr references

theemadnes:
fixed erroneous field name references to cloud run when they were actually just part of GCE (instance ID + default service account) and also updated protobufs to use `uint64` instead of `string` for instance ID which was causing errors.


xtineskim (approved):
LGTM! 👍

PR #461, opened 2023-01-13
Enable autoscaling on GKE stateful DBs

aburhan:
Updated scaling to true for pool db - gmp deploys on this node-pool and causes scaling issues


bourgeoisor (approved):
LGTM

PR #492, opened 2023-02-14
Adapt names to reviewed Kueue documentation

rbarberop:
The Kueue documentation has been updated recently.

This PR uses the same ResourceFlavor names than in documentation.

It also gives different names to jobs generated for both teams, so the output of the commands in the tutorial are more descriptive.


rbarberop:
FYI @alculquicondor @alizaidis

PR #516, opened 2023-03-09
GKE Stateful Kafka Tutorial

aburhan:


bourgeoisor (approved):
LGTM, with the caveat that there may be further PRs to polish and fix any outstanding issues before we publish

PR #530, opened 2023-03-27
cleanup: add renovate rule to skip springboot upgrade

Shabirmean:
#### Description
- This change is related to https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/410
- Renovate tries to update springboot-starter version which relies on Java17 
- Our CI doesn't support Java 17 yet, thus we have to wait until that change is done.
- Until then we can skip it.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -33,5 +33,11 @@
   },
   "nuget":{
     "addLabels": ["lang: dotnet"]
-  }
+  },
+  "packageRules": [
+    {
+      "matchPackageNames": ["org.springframework.boot:spring-boot-starter-parent"],
+      "allowedVersions": "<3.0.0"
+    }
+  ]
 }


bourgeoisor:
This seems overkill, I think it'd be better if we simply close #410 instead (and open a ticket to fix this)
Once closed, Renovate doesn't reopen for that version.

bourgeoisor:
What do you mean by "Our CI doesn't support Java 17 yet" exactly?

Afaik this has nothing to do with the CI, rather we just need to use a different image for this which supports Java 17+: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/quickstart/java/Dockerfile

Shabirmean:
Closing this PR as suggested in comment: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/530#issuecomment-1485439161 and also closing related PR: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/410

PR #572, opened 2023-04-25
fix: upgrade to latest pubsub client lib (2.16)

### LINKED ISSUE #284: Bump PubSub
_Major_ bump in our dependency google-cloud-pubsub to v2 will require more work than what renovate bot PRs are showing. 
https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/280

minherz:
Fixes #284

### DIFF
--- a/.github/workflows/pubsub-ci.yml
+++ b/.github/workflows/pubsub-ci.yml
@@ -1,0 +1,21 @@
+name: pubsub-ci
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/pubsub-ci.yml'
+      - 'cloud-pubsub/**'
+  pull_request:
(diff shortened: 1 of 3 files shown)


bourgeoisor:
@minherz with such a big jump in versions I suspect it may not "just work" like that, make sure that you can run `docker build` successfully without complaints, and see what logs `docker run` gives you. If it's "unauthorized", that's fine (it can't connect to PubSub), but if you get python errors, then there may be code changes necessary

minherz:
Need to modify the code so it will be compatible with pubsub version 2.16.
Current modification needs to retrieve a "default" project ID in order to run.

bourgeoisor (approved):
LGTM

Issue #571, opened 2023-04-24
Revise samples for autopilot networking tutorial


melissakendall:
Hello,

This is a request to revise /kubernetes-engine-samples/tree/main/autopilot/basic-networking-bank-of-anthos as follows:
- Split ingress.yaml into seperate manifests by KIND, which helps with formatting the tutorial that will use those manifests
- Format variable names as caps (i.e. CERTIFICATE_NAME) to prepare for variable insertion in the tutorial that will use this variables and remove "bank-of-anthos" for each name, which is difficult for maintenance and logical understanding of each variable (if everything is named "bank-of-anthos", its hard to track names through the tutorial)
- Insert Terraform manifests so they are not hard-coded in the tutorial that will use this folder (optional, will discuss with team)

I request permission to create a branch to make these changes.

Thank youm
-Melissa

minherz:
Hi @melissakendall ,
Thank you for your input. We agree that clarity of the source code is important. Feel free to [fork this repo][1] and submit the changes that you see fit via a [pull request][2]. You do not need special permissions to do it. If you contribute to this repo first time, please familiarize with the [CONTRIBUTING][3] guide.

[1]: https://docs.github.com/en/get-started/quickstart/fork-a-repo
[2]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
[3]: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/1cf2f30520e0615715fc07f77c9a1548dc07c00d/CONTRIBUTING.md

bourgeoisor:
Alternatively to splitting the manifest, you can instead use region tags surrounding the lines that you want to integrate in the docs (example: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/quickstart/deployment.yaml#L16)

bourgeoisor:
@melissakendall if you put "Fixes #123" with the issue number in the description of a PR, it'll tie them together and closing the PR will also close the issue -- I will reopen and do that now!

PR #573, opened 2023-04-25
Autopilot networking tutorial manifest changes

melissakendall:
Fixes https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/571

Summary of changes:
- Remove any references to "simple" or "basic" as those words are not advised as per Google developer style guide
- Removed "bank of anthos" as the tutorial that uses these manifests does not modify the standard Bank of Anthos app, and cloning 2 repos was unnecessary. Best to keep everything in one repo, and make it simpler.
- Added Terraform snippets. NOTE: I did not create these, and have limited ability to test them. I am relying on SWEs to identify if there are any errors.
- Changed object naming to make it easier to track objects (instead of everything being named "bank-of-anthos")

I'd love to squash my commits, but I appear unable to do that through GitHub UI. I think I can do it when I merge the PR though.


autopilot/networking-tutorial/dns.tf
```diff
@@ -0,0 +1,69 @@
...
+variable "name" {
+  type        = string
+  description = "Name of resources"
+  default     = "networking-tutorial"
+}
+
+data "google_client_config" "current" {}
+
+resource "google_compute_global_address" "default" {
+  name = var.name
```
melissakendall, line 35: Does this mean that the global IP address that gets created will have the standard name "networking-tutorial"? If so I should change this to match the IP address referenced in kubernetes-manifests.yaml.

melissakendall:
Incorporated revisions from a GKE subject matter expert who tested the Terraform snippets.

bourgeoisor:
@melissakendall `networking-tutorial` seems a bit vague as the directory name. I'm "okay" with it but I'm worried that eventually we may have samples for another "networking tutorial" and naming may become confusing. Is there an alternative we could use?

melissakendall:
Fair enough. "basic networking tutorial"?

I just dont want to add "bank of anthos" because the goal of the tutorial is to resolve a concern from customers that setting up an AP cluster is easy but adding all the networking requirements is difficult. To me, that means that adding BoA would add too much complexity (and the customer would have to checkout an additional repo)

melissakendall:
> @melissakendall `networking-tutorial` seems a bit vague as the directory name. I'm "okay" with it but I'm worried that eventually we may have samples for another "networking tutorial" and naming may become confusing. Is there an alternative we could use?

Fair enough. "basic networking tutorial"?

I just dont want to add "bank of anthos" because the goal of the tutorial is to resolve a concern from customers that setting up an AP cluster is easy but adding all the networking requirements is difficult. To me, that means that adding BoA would add too much complexity (and the customer would have to checkout an additional repo)

bourgeoisor:
What's the name of the tutorial? Maybe we can play on that!
Agreed with BoA being complex for more intro tutorials that don't require multiple services, keeping it simple works wonders

melissakendall:
> What's the name of the tutorial? Maybe we can play on that! Agreed with BoA being complex for more intro tutorials that don't require multiple services, keeping it simple works wonders

Currently: 
Configure networking for a basic production cluster

But I am not sold on "basic" or "production"

bourgeoisor (approved):
Still unsure about the directory name `networking-tutorial` but I don't feel super strongly about it. LGTM! Let me know when / if it's time to merge this.

melissakendall:
Ready to submit at any time!

PR #608, opened 2023-05-17
fix: Upgrade pubsub client library to 2.*

### LINKED ISSUE #284: Bump PubSub

minherz:
Upgrade pubsub client library version from 0.25 to 2.17.
Modify code to use new interfaces of the version 2.
Update the tags for the docker images to run the new code.

Fixes #284


cloud-pubsub/main.py
```diff
@@ -8,37 +8,43 @@
...
-    """Process received message"""
-    print("[{0}] Processing: {1}".format(datetime.datetime.now(),
-                                         message.message_id))
-    time.sleep(3)
-    print("[{0}] Processed: {1}".format(datetime.datetime.now(),
-                                        message.message_id))
+
+    # read default project ID
+    _, project_id = auth.default()
+    subscription_id = 'echo-read'
```
anguillanneuf, line 22: question: is hard-coding a subscription id to be `echo-read` required? i'm in favor of changing it to `${TOPIC_ID}-sub` to let users know it is a subscription paired with a specific topic, with a comment saying this subscription is attached to this topic

cloud-pubsub/main.py
```diff
@@ -8,37 +8,43 @@
...
+    subscription_path = subscriber.subscription_path(
+        project_id, subscription_id)
+
+    def callback(message: pubsub_v1.subscriber.message.Message) -> None:
+        """Process received message"""
+        print(f"Received message: ID={message.message_id} Data={message.data}")
+        message.ack()
+        print(f"[{datetime.datetime.now()}] Processing: {message.message_id}")
+        time.sleep(3)
+        print(f"[{datetime.datetime.now()}] Processed: {message.message_id}")
```
anguillanneuf, line 34:
```suggestion
        """Process received message"""
        print(f"Received message: ID={message.message_id} Data={message.data}")
        print(f"[{datetime.datetime.now()}] Processing: {message.message_id}")
        # Simulate message processing with sleep
        time.sleep(3)
        message.ack()
```

you should always ack messages after they have been processed not before because ack'ed messages will be gone (generally speaking).

call `message.ack()` last in this callback functionn.

anguillanneuf (changes requested):
mostly LGTM. i don't see a `main-test.py`, will you be adding a test?

cloud-pubsub/main.py
minherz: Yes, it is. It follows the tutorial that instructs to create a fixed subscription name.

cloud-pubsub/main.py
minherz: I ported the existing code to the new interface. I do not know if separating the data reading and processing has a meaning. But I prefer to avoid changing the logic around that. Also one can argue that reading the data and processing the data can be separate tasks.

bourgeoisor (approved):
LGTM from a repo standpoint

cloud-pubsub/main.py
anguillanneuf: it's important to show ack() after processing, not before. as mentioned, acking before messages are successfully processed can cause message loss.

minherz: done

anguillanneuf (approved):
Approve but I still don't see a test. I don't see why not.

PR #642, opened 2023-06-09
Fix typo in the PubSub sample

bourgeoisor:
This PR fixes a typo in the PubSub samples, where `feature` should be `future`, to match https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.subscriber.futures.StreamingPullFuture

### DIFF
--- a/cloud-pubsub/main.py
+++ b/cloud-pubsub/main.py
@@ -33,13 +33,13 @@ def main():
         print(f"[{datetime.datetime.now()}] Processed: {message.message_id}")
         message.ack()
 
-    streaming_pull_feature = subscriber.subscribe(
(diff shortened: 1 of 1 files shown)


minherz (approved):
nit: My guess is that it was added here to demonstrate a _feature_. Renaming it does not make any sense. I suggest to remove the suffix (`_feature`) instead.

bourgeoisor:
The official docs for that method (`.subscribe()`) seems to disagree: https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.subscriber.client.Client#google_cloud_pubsub_v1_subscriber_client_Client_subscribe

It "returns a future"

PR #666, opened 2023-06-16
refactor: update app entity

thiagotnunes:
Updates the application example entity from Singers to Players to better align with existing GKE samples with Cloud Spanner.

Fixes PostgreSQL dialect query in the application to use PG style parameters.


bourgeoisor (approved):
LGTM!

PR #673, opened 2023-06-20
Update T5 samples to use Artifact Registry and tune the load test

shannonxtreme:
- Change the images to live in Artifact Registry instead of Container Registry.
- Use only one repo `models` for all images
- For the load generator, increase spawn rate and max users to improve chance of triggering a scaleup
- For HPA, reduce the latency target to improve change of triggering scaleup


ai-ml/t5-model-serving/model/cloudbuild.yaml
```diff
@@ -36,7 +36,7 @@ substitutions:
   _MODEL_NAME: t5-small
   _MODEL_VERSION: "1.0"
   _MACHINE: cpu
-  _MODEL_IMAGE: gcr.io/${PROJECT_ID}/models/${_MODEL_NAME}:${_MODEL_VERSION}-${_MACHINE}
+  _MODEL_IMAGE: ${_LOCATION}-docker.pkg.dev/${PROJECT_ID}/models/${_MODEL_NAME}:${_MODEL_VERSION}-${_MACHINE}
```
bourgeoisor, line 39: This uses an env for the location, but everywhere else in this PR it's hardcoded as `us-central1`, is that intentional?

NimJay: CC: @shannonxtreme (in case your missed this question we had) :)

shannonxtreme: It was intentional yes, the tutorial passes the location alongside the MODEL_NAME and MODEL_VERSION variable susbstitutions

TheRealSpaceShip: Need to set a default value for `_LOCATION` substitution, same as `_MODEL_NAME`, `_MODEL_VERSION`, etc.
```yaml
_LOCATION: us-central1
```

shannonxtreme: Done. @NimJay I'm not sure whether my push muddled up someone else's commit though, there's a fifth commit there now

bourgeoisor (approved):

PR #721, opened 2023-07-14
feat: samples for upcoming Airflow tutorial

TheRealSpaceShip:
These are Airflow tutorial-related files.

The tutorial will describe the process of deploying the Airflow platform on GKE Autopilot with next features:
 - Loading DAG files from the GCS bucket by using the GCSFuse CSI driver and mounting volume to Airflow pods. 
 - Configured Ingress with SSL Google-managed certificate. 
 - SSO authentication with Google via OpenID Connect.
 - Monitoring of Airflow with GMP (Google-managed Prometheus)

Also, PR contains a simple Python script to generate secrets for Airflow (Web Server Secret Key and Fernet key) and a simple example of a DAG file.


bourgeoisor:
@TheRealSpaceShip Is there an ETA for this tutorial?

PR looks good at first glance!

TheRealSpaceShip:
@shannonxtreme Could you please provide ETA for this tutorial?

shannonxtreme:
Im not sure @TheRealSpaceShip it's not on my backlog

batch/airflow/airflow-sa.yaml
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: airflow
+  namespace: airflow
+  annotations:
+    iam.gke.io/gcp-service-account: airflow-dags-reader@lustrous-baton-363720.iam.gserviceaccount.com
```
bourgeoisor, line 21: Is this hardcoded to somewhere? (`lustrous-baton-363720`)

batch/airflow/values.yaml
```diff
@@ -0,0 +1,294 @@
...
+      ephemeral-storage: 1Gi
+    requests:
+      cpu: 250m
+      memory: 512Mi
+      ephemeral-storage: 1Gi
+
+postgresql:
+  enabled: false
+
+dags:
```
bourgeoisor, line 272: Should these be commented out?

batch/airflow/example_dag.py
```diff
@@ -0,0 +1,70 @@
+#
```
bourgeoisor, line 1: Is this python script running inside of a container? Do we need a Dockerfile checked in?

batch/airflow/secrets/requirements.txt
```diff
@@ -0,0 +1 @@
+cryptography
```
bourgeoisor, line 1: Would you mind pinning this to a version?

batch/airflow/airflow-sa.yaml
TheRealSpaceShip: It's leftover from testing. I'll fix it.

batch/airflow/values.yaml
TheRealSpaceShip: No, postgresql is enable by default but in tutorial I'm using CloudSQL instance istead. And dags is to configure persistence volume that will be mounted to Airflow pods.

batch/airflow/example_dag.py
TheRealSpaceShip: This is [DAG](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html). Airflow runs it inside of special pod (mount volume with this script which should be placed in GCS bucket). So, no need of Dockerfile.

melissakendall:
Please make sure to add Region tags so we can embed these in the docs (we have been inconsistent with doing this, but it is technically a requirement)

See comment here: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/804#pullrequestreview-1624728019

bourgeoisor (approved):
LGTM.

Let me know when I can merge.

Please also create a follow-up PR once the tutorial is published to add the URL of the tutorial in the README.

minherz:
@TheRealSpaceShip please rebase and merge

TheRealSpaceShip:
@minherz I have removed `example_dag.py` file because I took this example from the official Airflow repository and it has a different license header that fails `header-check`. I propose to mention in a tutorial that the user needs to download this file with a command like `curl -L https://raw.githubusercontent.com/apache/airflow/main/airflow/example_dags/example_bash_operator.py -o example_dag.py`

TheRealSpaceShip:
@bourgeoisor this PR is done and ready to be merged.

PR #762, opened 2023-07-31
Ignore Renovate deps for security/language-vulns

bourgeoisor:
This PR:
- Reverts an unintentional dependency bump on `security/language-vulns`
- Modifies the Renovate config to ignore that directory moving forward

This is due to the fact that these samples _intentionally_ hosts outdated images for demo purposes.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -33,5 +33,6 @@
   },
   "nuget":{
     "addLabels": ["lang: dotnet"]
-  }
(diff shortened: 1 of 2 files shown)


NimJay (approved):
Looks good! Approved.
Thanks, @bourgeoisor, for fix and explanation (i.e., "intentionally hosts outdated images for demo purposes").

PR #818, opened 2023-08-18
Fix wi-secrets image build

bourgeoisor:
This fixes a broken build of wi-secrets by generating and checking in a Go module checksum, rather than relying on unversioned package names (non-deterministic).

This is the broken build where I noticed this: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/actions/runs/5883405124/job/15955997993?pr=805


NimJay (approved):
Thanks for fixing this.
I'm wondering why we used `go get` inside the Dockerfile in the first place for this `secretmanager` library. I'm going to assume it wasn't intentional.
So approved!

PR #829, opened 2023-08-24
replace legacy bigquery inserts with bigquery storage API

aburhan:


aburhan:
Hi @bourgeoisor, I made an update can you PTAL ?

bourgeoisor (approved):

Issue #826, opened 2023-08-22
fix HTTP retries on whereami


theemadnes:
sometimes (~20% of time) when running in a service mesh, the app container won't be able to call the metadata service because it is briefly blocked by the sidecar - thought this was fixed by adding some retry logic, but the retry logic is flawed, so need to fix that.

theemadnes:
ugh so i thought i "fixed" it but retry logic is way too long - it also impacts downstream calls, not just calls to mds. example: ```$ time curl http://34.136.219.95 -v
*   Trying 34.136.219.95:80...
* TCP_NODELAY set
* Connected to 34.136.219.95 (34.136.219.95) port 80 (#0)
> GET / HTTP/1.1
> Host: 34.136.219.95
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: Werkzeug/2.3.7 Python/3.11.4
< Date: Fri, 08 Sep 2023 02:37:06 GMT
< Content-Type: application/json
< Content-Length: 565
< Access-Control-Allow-Origin: *
< Connection: close
< 
{"backend_result":null,"cluster_name":"autopilot-cluster-us-central1","gce_instance_id":"3292785587396495807","gce_service_account":"gateway-multicluster-01.svc.id.goog","host_header":"34.136.219.95","metadata":"frontend","node_name":"gk3-autopilot-cluster-us-centr-pool-2-e821940b-s8x5","pod_ip":"10.127.0.152","pod_name":"whereami-frontend-bfff7cb89-c94mc","pod_name_emoji":"\u2b06","pod_namespace":"whereami-http-frontend","pod_service_account":"whereami-frontend","project_id":"gateway-multicluster-01","timestamp":"2023-09-08T02:34:56","zone":"us-central1-c"}
* Closing connection 0

real    2m10.251s
user    0m0.017s
sys     0m0.007s```

theemadnes:
in the above example the call that populates `backend_result` times out, but the retry logic takes over 2 minutes. will fix.

PR #830, opened 2023-08-24
826 http retries whereami

theemadnes:
implements more robust retries for when the container starts and attempts to contact the metadata service.

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/826

cc: @gbrayut


theemadnes:
also incremented base python-slim image and `curl` version

bourgeoisor (approved):

PR #841, opened 2023-08-31
Update kubernetes-manifests.yaml

### LINKED ISSUE #840: Tutorial uses wrong value
https://cloud.google.com/kubernetes-engine/docs/tutorials/configure-networking

dns.tf  specifies a google_compute_gloval_address with name var.name, which is associated with the variable name (networking-tutorial)

Later on, in kubernetes-manifests.yaml, the Ingress annotation ingress-global-static-ip-name is "networking-static-ip"

I believe it should, instead, be "networking-tutorial"

I can make a PR for this.

melissakendall:
Update variable name

Fixes #840 

Let me know if I did this wrong or if you see anything else.

### DIFF
--- a/autopilot/networking-tutorial/kubernetes-manifests.yaml
+++ b/autopilot/networking-tutorial/kubernetes-manifests.yaml
@@ -70,7 +70,7 @@ metadata:
   annotations:
     networking.gke.io/managed-certificates: networking-managed-cert
     networking.gke.io/v1beta1.FrontendConfig: networking-fc
-    kubernetes.io/ingress.global-static-ip-name: networking-static-ip
+    kubernetes.io/ingress.global-static-ip-name: networking-tutorial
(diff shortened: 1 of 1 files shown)


NimJay (approved):
Thank you for this fix, @melissakendall!
Your fix looks right to me since the [`name` Terraform variable defaults to `networking-tutorial`](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/059f6e0088af552cd01f55f8da691bf322732b71/autopilot/networking-tutorial/dns.tf#L29):
```
variable "name" {
  type        = string
  description = "Name of resources"
  default     = "networking-tutorial"
}
```
And the `google_compute_global_address` Terraform resource uses that `name` variable as its name.
```
resource "google_compute_global_address" "default" {
  name = var.name
}
```

PR #857, opened 2023-09-03
Update golang to 1.21 in all samples

bourgeoisor:
This PR updates golang to `1.21` (from `1.19`) in all samples, tidying up the modules along the way.


NimJay (approved):
Thank you for this update, @bourgeoisor!
Looks good.

PR #909, opened 2023-09-27
Update helloweb-hpa.yaml

melissakendall:
A Google stakeholder instructed that the API version is deprecated and we should update it

Deprecation notice: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126

Related tutorial: https://cloud.google.com/kubernetes-engine/docs/tutorials/autoscaling-metrics#step4


bourgeoisor (approved):
LGTM!

PR #887, opened 2023-09-18
Fix t5 container build

bourgeoisor:
This fixes the T5 sample container build, which was failing on the main branch.

Note that the container build takes ~8 minutes to complete from scratch.

PR #942, opened 2023-10-09
fix: Fix client-app and model in t5-model tutorial

TheRealSpaceShip:
After #887 version of the API of the fast-dash library has changed and the current version of the application does not work. 

- Update client-app to use actual API of fast-dash lib
- Freeze dependencies for model
- Update model's and client-app's README files

### DIFF
--- a/ai-ml/t5-model-serving/client-app/src/app.py
+++ b/ai-ml/t5-model-serving/client-app/src/app.py
@@ -29,7 +29,7 @@ LANG_MAP = {
   "en": "English",
   "fr": "French",
   "de": "German",
-  "es": "Spanish",
+  "ro": "Romanian",
 }
 
 logger.info(f"Model prediction: {PREDICTION_URL}")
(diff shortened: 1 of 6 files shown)


ai-ml/t5-model-serving/client-app/src/app.py
```diff
@@ -40,15 +40,18 @@ def text_to_text_function(text: str = "Hello World", from_lang=LANG_MAP, to_lang
   resp = requests.post(PREDICTION_URL, json=payload, headers=headers)
   if resp.status_code == 200:
     content = resp.json()
-    return content.get("text")
+    resp = content.get("text")
   else:
-    return "Oops, something went wrong!"
+    resp = "Oops, something went wrong!"
+
+  return resp
```
bourgeoisor, line 47: This code change doesn't change the flow, no?

bourgeoisor (approved):
LGTM

ai-ml/t5-model-serving/client-app/src/app.py
TheRealSpaceShip: The flow stayed the same except `text_to_text_function` should have only one return line at the end of the function. The function `text_to_text_function` is a callback passed to the FastDash constructor that calls [_infer_variable_names](https://github.com/dkedar7/fast_dash/blob/9a58f2d13cc8e578cdd1ee460f28d54ffc4bee74/fast_dash/utils.py#L391). As result `_infer_variable_names` is a root cause of this change.

bourgeoisor:
@TheRealSpaceShip let me know when this is good to merge from your point of view

TheRealSpaceShip:
@bourgeoisor PR is ready. Sorry for the additional commits. I retested the model + client-app to be sure that everything worked as planned.

bourgeoisor (approved):

PR #951, opened 2023-10-11
Remove unused k8s file

### LINKED ISSUE #557: Remove hpa-example from sample manifests
`gke-scheduled-autoscaler/k8s/deployment-example.yaml` uses "k8s.gcr.io/hpa-example" as the image. It's not used in any of our docs, so I think this should be ok to change. 

In https://github.com/kubernetes/release/issues/2924#issuecomment-1454301010, it's suggested that `hpa-example` is going to be removed in the near future. We should replace the image in this repo with either a forked version or a different image.

cc @bourgeoisor

bourgeoisor:
Fixes #557 

cc: @shannonxtreme


NimJay (changes requested):
Thanks for this clean-up.

Should we also remove [`gke-scheduled-autoscaler/k8s/hpa-example.yaml`](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/86241fa0540350eef39bd27f3d61a55f98b74b36/gke-scheduled-autoscaler/k8s/hpa-example.yaml)?

Also, I just noticed that the region tag for the following samples are identical:
* [`/gke-scheduled-autoscaler/k8s/scheduled-autoscaler/hpa-example.yaml`](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/86241fa0540350eef39bd27f3d61a55f98b74b36/gke-scheduled-autoscaler/k8s/scheduled-autoscaler/hpa-example.yaml)
* [`gke-scheduled-autoscaler/k8s/hpa-example.yaml`](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/86241fa0540350eef39bd27f3d61a55f98b74b36/gke-scheduled-autoscaler/k8s/hpa-example.yaml)

bourgeoisor:
@NimJay no, `hpa-example.yaml` is used here: https://cloud.google.com/kubernetes-engine/docs/tutorials/reducing-costs-by-scaling-down-gke-off-hours#deploy_the_example_application

I also don't want to modify any region tags in this PR since it'd require changing the published tutorial source code as well.

gke-scheduled-autoscaler/k8s/deployment-example.yaml
NimJay, line 1: Good find! (pointing out that 

Also, it looks like this `deployment-example.yaml` is also used by [that tutorial](https://cloud.google.com/kubernetes-engine/docs/tutorials/reducing-costs-by-scaling-down-gke-off-hours#deploy_the_example_application). 
```
kubectl apply -f ./k8s
```

bourgeoisor: Oh good catch. I had searched for the file name without realizing that command applied everything at once. Seems like we won't be able to simply remove it after all.

PR #952, opened 2023-10-11
Spotahome Redis operator - fixes

ganochenkodg:
Resolving guide comments after tech-review


bourgeoisor (approved):

bourgeoisor:
@ganochenkodg lemme know when this is good to merge in

databases/redis-spotahome/terraform/gke-autopilot/versions.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = "~> 5.0"
```
bourgeoisor, line 5: Note that Renovate will try to automatically upgrade these again as soon as this PR merges in.

Is there any reason latest TF provider is not compatible?

If yes and we can't go around it, you might have to either a) update the CI job such that it fails if a new PR upgrading to 5 occurs again or b) tweak the Renovate config to permanently set this version to 4

bourgeoisor (changes requested):

databases/redis-spotahome/terraform/gke-autopilot/versions.tf
ganochenkodg: Yep, when I use ~> 5.0 i get failed:
Screenshot 2023-10-13 at 00 51 59

Other people from the team also got the same issue, so i decided to downgrade the version (it used to work with ~> 4.0)

bourgeoisor: Would you mind updating the databases/ CI with an additional job for this "redis-spotahome" directory, with steps for each TF to do `terraform init` + `terraform validate`?

That way when Renovate creates another PR to bump to v5 of the provider, we'll immediately see those checks fail and won't merge it accidentally

ganochenkodg: Added CI steps

bourgeoisor (approved):

PR #977, opened 2023-10-23
chore(proc): update renovate configuration

minherz:
use json5 format in the config file.
schedule updates to weekly on Monday.
group updates per language.
group terraform providers together.
disallow automatic major version updates


.github/renovate.json5
```diff
@@ -0,0 +1,91 @@
...
+  "semanticCommits": true,
+  "pip-compile": {
+    "enabled": true,
+    "fileMatch": ["(^|/)requirements\\.in$"],
+    "lockFileMaintenance": {
+      "enabled": true
+    },
+  },
+  // pip-compile handles requirements.txt updates.
+  "pip_requirements": {
```
muncus, line 31: We should not disable this manager here, because there are many `requirements.txt` files that are not generated by pip-compile in this repo.

.github/renovate.json5
```diff
@@ -0,0 +1,91 @@
...
+  "labels": ["dependencies"],
+  "dependencyDashboardLabels": ["type: process"],
+  "dockerfile": {
+    "enabled": false
+  },
+  "semanticCommits": true,
+  "pip-compile": {
+    "enabled": true,
+    "fileMatch": ["(^|/)requirements\\.in$"],
+    "lockFileMaintenance": {
```
muncus, line 26: This is the default, and can safely be removed.

muncus (changes requested):

.github/renovate.json5
```diff
@@ -0,0 +1,91 @@
...
+      "matchPackagePatterns": [
+        "org.springframework.boot"
+      ],
+      "matchCurrentVersion": "!/.*[SNAPSHOT | M]/",
+      "allowedVersions": "<3",
+      "description": "@akitsch: Spring Boot V3 requires Java 17"
+    },
+    {
+      "matchCategories": ["terraform"],
+      "matchManagers": ["google"],
```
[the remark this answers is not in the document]
minherz: Thank you! I was trying to figure out how to express it and.. 😕

.github/renovate.json5
```diff
@@ -0,0 +1,91 @@
...
+      "groupName": "Jetty packages"
+    },
+    {
+      "matchCategories": ["java"],
+      "matchPackagePatterns": [
+        "org.springframework.boot"
+      ],
+      "matchCurrentVersion": "!/.*[SNAPSHOT | M]/",
+      "allowedVersions": "<3",
+      "description": "@akitsch: Spring Boot V3 requires Java 17"
```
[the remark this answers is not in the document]
minherz: Would it just prevent major upgrades for spring at all including the dashboard? Ideally I do not want them to be there either because we aren't going to drop Java8 support in near future and I want to avoid any confusion.

.github/renovate.json5
```diff
@@ -0,0 +1,91 @@
...
+    {
+      "matchDepTypes": ["devDependencies"],
+      "automerge": true
+    },
+    {
+      "matchCategories": ["java"],
+      "matchPackagePatterns": [
+        "io.grpc:(?:protoc-gen-)?grpc-*"
+      ],
+      "groupName": "gRPC packages"
```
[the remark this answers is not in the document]
minherz: copied-n-pasted from the old version. Makes sense to keep it at the language level and then to compartmentalize if necessary.

.github/renovate.json5
```diff
@@ -0,0 +1,91 @@
+{
+  "extends": [
+    "config:recommended",
+    ":approveMajorUpdates",
+  ],
+  // submit on Monday midnight (UTC)
```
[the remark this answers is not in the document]
minherz: :slightly_smiling_face:

.github/renovate.json5
minherz: I am (almost) sure that I migrated it from the original. Based on the comment I was under impression that it enforces handling of the .txt file to the `pip-compile`. Am I wrong?

.github/renovate.json5
minherz: Looking at the [version] you composed for BoA, I see the difference is only in the `lockFileMaintenance` field.
Should it be removed in both locations?

[version]: https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/5c9539ea0b9e4e1e6830fcb64ad4b27429630087/.github/renovate.json5#L9-L12

muncus: apologies, i was unclear. The `lockFileMaintenance` stanza should be removed here. it is not present in the BoA variant.

.github/renovate.json5
muncus: BoA **exclusively** uses pip-compile to turn `requirements.in` into `requirements.txt`, so it is safe to disable this manager there. That is not the case for this repository - here we have a mix of "regular" and pip-compile requirements files.
I do not know if these two methods can peacefully coexist, but we certainly should not be disabling pip_requirements in this repo.

.github/renovate.json5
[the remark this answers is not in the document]
muncus: AIUI, it will not make it disappear from the Dependency Dashboard, because that is not something Renovate does. It will appear as "ignored or blocked", just as the current spring 3x update does (https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/234).
The current springboot 3x update is blocked because the PR was closed un-merged. This is the standard way to signal that renovate should ignore an upgrade. No configuration is needed beyond closing the PR without merging.

if it is really important to you to encode this springboot requirement, i'd recommend simplifying this stanza to something more like:
```
"matchCurrentVersion": ">=2.0.0, <3.0.0",
"allowedVersions": "<3.0.0"
```
This is more explicit that the goal is to preserve springboot 2x.

.github/renovate.json5
```diff
@@ -0,0 +1,64 @@
...
+    "fileMatch": ["(^|/)requirements\\.in$"],
+  },
+  "constraints": {
+    "python": "~=3.11.0"
+  },
+  "ignorePaths": [
+    "demos/**",
+    "security/language-vulns/**"
+  ],
+  "packageRules": [
```
muncus, line 34: Please re-add the per-language grouping rules. They should look like this:

```
    {
      "matchCategories": ["golang"],
      "matchUpdateTypes": ["minor", "patch", "digest", "lockFileMaintenance"],
      "groupName": "golang",
      "automerge": true
    },
```
one rule for each matchCategories: python,golang, java,js,dotnet,docker

Note that lockFileMaintenance is new, and i'm making corresponding PRs for the other repos now.

minherz: What is a reason for duplicated rules? There is already https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/977/files#diff-20ab1190d08a53a3012bc191ed56205c8f0ffb8fa1715e9d36ecfd1d2ea8a790R40-R43

muncus (approved):
Thanks, @minherz !

PR #982, opened 2023-10-23
feat: make a home for renovate config presets

muncus:
dee-platform-ops team preset, and docs on using renovate config presets.


renovate-configs/dee-platform-ops.json5
```diff
@@ -0,0 +1,86 @@
+// Shared Renovate Base Configuration for repos operated by the
+// @GoogleCloudPlatform/dee-platform-ops team.
```
NimJay, line 2: **Praise:** These comments add a lot of clarity! I appreciate that you refer to `@GoogleCloudPlatform/dee-platform-ops`.

renovate-configs/dee-platform-ops.json5
```diff
@@ -0,0 +1,86 @@
...
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+  ],
+  // Rate limits, to protect maintainers (and CI) from PR floods.
+  "prHourlyLimit": 10,
+  "prConcurrentLimit": 30,
```
NimJay, line 10: **Thought:** I'm guessing this `"prConcurrentLimit": 30,` is _per repo_. That is, at any given time, we can have 30 Renovate PRs in Bank of Anthos, 30 in Online Boutique, 30 in GKE samples, etc.
I'd lower this number to about 10 — since grouping by language also reduces the total repo count.

renovate-configs/README.md
NimJay, line 1: This README.md is helpful!
Especially:
```
{
  "extends": [
    "github>GoogleCloudPlatform/kubernetes-engine-samples//renovate-configs:purple.json5",
  ],
}
```

Good to know that this is how we share a single config across multiple repos.

NimJay, line 1: This README.md is helpful!
Especially:
```
{
  "extends": [
    "github>GoogleCloudPlatform/kubernetes-engine-samples//renovate-configs:purple.json5",
  ],
}
```

Good to know that this is how we share a single config across multiple repos.

renovate-configs/dee-platform-ops.json5
```diff
@@ -0,0 +1,86 @@
...
+    "type: process",
+  ],
+  "packageRules": [
+    // Label PRs based on the languages they contain
+    // category ids are listed at https://docs.renovatebot.com/modules/manager/
+    {"matchCategories": ["python"], "addLabels": ["lang: python"]},
+    {"matchCategories": ["java"], "addLabels": ["lang: java"]},
+    {"matchCategories": ["golang"], "addLabels": ["lang: go"]},
+    {"matchCategories": ["js"], "addLabels": ["lang: nodejs"]},
+    {"matchCategories": ["dotnet"], "addLabels": ["lang: dotnet"]},
```
NimJay, line 30: **Not action needed:** Just so you're aware. We also have a:
* one-off Ruby sample inside the GKE samples repo: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/quickstart/ruby/Gemfile.
* one-off PHP sample inside the GKE samples repo: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/guestbook/php-redis/composer.json

Sounds like Renovate [supports PHP](https://docs.renovatebot.com/php/) and [supports Ruby](https://docs.renovatebot.com/ruby/).
But I don't think we need to be that thorough. Up to you.

NimJay (approved):
This looks good to me!
Thanks again, @muncus 👏 , for driving this renovation of Renovate (🥁 ⚪ ).

renovate-configs/README.md
bourgeoisor, line 1: I'd prefer hiding this directory of view (so, in the `.github/` directory) to not mistake it for user-facing samples

NimJay: +1 to stuffing `renovate-config` inside `.github`.
(Reminder to ourselves: ensure `README.md ` is updated to this new directory before merging.)

bourgeoisor: Clarifying that I meant `.github/renovate-blah-blah/` and not directly under `.github/` :)

renovate-configs/dee-platform-ops.json5
muncus: Yes, your understanding is correct, this is a per-repo number.
I admit this number is pretty arbitrary, and I appreciate your justification.
In regular practice, we should come nowhere near this number, because of the grouping, as you noted.

As I see it, the consequence for exceeding this number is a potentially long delay before the update PRs are created, since we're only running once each week. However, there are 2 reasons I think its ok to reduce this somewhat:
1) the dependency dashboard issue will still show rate-limited updates, and will create them when requested.
2) This is a base configuration, and each repo can override it.

What do you think about 15 as a starting point? we may want to tune it lower for OB.

renovate-configs/dee-platform-ops.json5
muncus: I was planning to do this in a future iteration - I'll add it as future work to the proposal so we don't lose track of it.

Another benefit to using a shared config is we can update this in one place, and see updates in all our repos 😄

renovate-configs/README.md
muncus: This is a great point, and I've made the change. Thanks!

muncus:
Thank you @NimJay and @bourgeoisor for your insightful comments! I believe i've addressed them all, PTAL.

renovate-configs/dee-platform-ops.json5
NimJay: 15 sounds good!

NimJay (approved):
Thanks for addressing all of our comments. Approved! 

This pull-request will also fix https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/985 (missing comma).

PR #999, opened 2023-10-31
feat(renovate): use dee-platform-ops renovate preset

muncus:
updates docs on how to use

Theres something odd going on with `ignorePaths` from one of the presets. It was overriding the paths in the in-repo config. so i've disabled that ignore preset for now while I continue to investigate the config merging.


.github/renovate.json5
```diff
@@ -64,31 +29,6 @@
       "groupName": "Spring Boot upgrades for v2",
       "description": "@akitsch: Spring Boot V3 requires Java 17"
     },
```
NimJay, line 31: **Thought:** I wonder if it's worth moving this to the global preset.
Point of Sale uses Java Spring Boot (see [related search in Point of Sale](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fpoint-of-sale%20spring&type=code)).

.github/renovate-configs/README.md
```diff
@@ -14,7 +14,7 @@ follows (the double slash is intentional!) :
 ```
 {
   "extends": [
-    "github>GoogleCloudPlatform/kubernetes-engine-samples//.github/renovate-configs:purple.json5",
+    "github>GoogleCloudPlatform/kubernetes-engine-samples//.github/renovate-configs/purple.json5",
```
NimJay, line 17: **Question:** Is there a reason that you suspect the `:` (colon) is an issue?
The [docs you shared earlier](https://docs.renovatebot.com/config-presets/#:~:text=GitHub%20with%20preset%20name%20(JSON5)) indicate that we need to use `:` (colon) for JSON 5.

NimJay (approved):
Approved! Feel free to merge.
My comments are mostly just curiosities — don't let them slow your progress.

.github/renovate-configs/README.md
NimJay: NVM, I just found [this example](https://github.com/syself/cluster-api-provider-hetzner/blob/e6128520aa716acb3a2b75c19e3e620b7e3e13c6/.github/renovate.json5#L10).

I'm likely misinterpreting the Renovate docs. 😕

muncus: Yeah, this is a consequence of renovate's peculiar syntax for the reference - apparently if you're [using a subdirectory path (indicated by `//`), the colon syntax is not supported](https://docs.renovatebot.com/config-presets/#github:~:text=anything%20in%20the%20form%20provider%3Eowner/repo//path/to/file%3Asubsubpreset%20is%20not%20supported.) and throws an error, but the full file path works just fine 🤷🏼

.github/renovate.json5
muncus: This is a good callout, i'll add it to my doc.
Given that springboot 3 -> 4 is likely to also be disruptive, we probably want a way to say "no major updates for springboot".

PR #1001, opened 2023-11-01
Add missing region tags to files used in tutorials

bourgeoisor:
This PR adds some region tags that were missing from files used in various tutorials. Licence headers have also been retroactively added where missing (in those files that were missing region tags, if applicable).


quickstarts/languages/nodejs/index.js
```diff
@@ -1,3 +1,18 @@
+// Copyright 2019 Google LLC
```
NimJay, line 1: **Praise:** You even dug up the correct year of creation (of each file). It's a nit, but appreciate the attention to detail!

PR #1002, opened 2023-11-02
Fix issue in MySQL StatefulSets example

NimJay:
* See Google-internal Buganizer: [b/304723700](https://b.corp.google.com/issues/304723700).

### DIFF
--- a/gke-stateful-mysql/kubernetes/prepare-for-ha.yaml
+++ b/gke-stateful-mysql/kubernetes/prepare-for-ha.yaml
@@ -31,6 +31,8 @@ spec:
         app: prepare-three-zone-ha
     spec:
       affinity:
+        # Tell Kubernetes to avoid scheduling a replica where there is
+        # already a replica with the label "app: prepare-three-zone-ha".
         podAntiAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
           - labelSelector:


gke-stateful-mysql/kubernetes/prepare-for-ha.yaml
```diff
@@ -31,6 +31,8 @@ spec:
         app: prepare-three-zone-ha
     spec:
       affinity:
+        # Tell Kubernetes to avoid scheduling a replica in a zone where there
+        # is already a replica with the label "app: prepare-three-zone-ha"
         podAntiAffinity:
```
NimJay, line 36: Let me know if I should remove this comment.
I personally found this chunk difficult to interpret, so I added a comment.
See [K8s documentation about podAntiAffinity here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#:~:text=In%20the%20following%20example%20Deployment%20for%20the%20Redis%20cache).

bourgeoisor (changes requested):
Wait hold on-- Can we make this change in the "new" sample location instead?

NimJay:
Good catch!
Fixed (just replicated the changes in the new sample location).

bourgeoisor (approved):

PR #1013, opened 2023-11-08
fix: Kafka Strimzi samples pod not starting

aburhan:
Resolved issues 
Kafkacat pod not starting due to secretName "my-user" not found 

b/309011900


streaming/kafka-strimzi/manifests/03-auth/my-user.yaml
```diff
@@ -34,34 +41,3 @@ spec:
           - Write
           - Read
         host: "*"
----
-apiVersion: v1
-kind: Pod
```
bourgeoisor, line 39: Is that Pod not necessary anymore?

bourgeoisor: @aburhan gentle ping

aburhan: yes this is duplicated in another manifest

bourgeoisor (approved):

PR #987, opened 2023-10-25
Restructure the repository by category (PR #1)

bourgeoisor:
This PR restructures the GKE samples by making sure that each sample lives in a subdirectory relating to its main point of interest (security, observability, networking, etc.)

To achieve this, we need to operate in 3 distinct steps:
- **Step 1**: duplicate each sample into its proper category sub-directory, and make sure that each triggers and workflows point to these new locations (this PR)
- **Step 2**: modify each GKE core tutorials utilizing these samples so they point the users to their new locations
- **Step 3**: finally, delete the old locations / duplicates of each samples

Fixes #688 

## What will this repo look like post-restructure?
Once each steps of the restructure are over, the root of this repo will look like this:
```
.github/
ai-ml/
autopilot/
batch/
cost-optimization/
databases/
demos/
management/
networking/
observability/
quickstarts/
security/
streaming/
windows/
.gitignore
LICENSE
README.md
```

## Which samples moved?
Some samples were already contained in category sub-directories. Those samples will not have to move.

This is the full list of moves contained in this PR:
```
cloud-pubsub -> databases
cloudsql -> (removed, this was already empty)
custom-metrics-autoscaling -> observability
gke-scheduled-autoscaler -> cost-optimization
gke-stateful-kafka -> streaming
gke-stateful-mysql -> databases
gke-stateful-postgres -> databases
gke-vpa-recommendations -> cost-optimization
guestbook -> quickstarts
hello-app -> quickstarts
hello-app-cdn  -> quickstarts
hello-app-redis -> quickstarts
hello-app-tls -> quickstarts
load-balancing -> networking
migrating-node-pool -> management
monitoring/distributed-tracing -> observability
network/dpv2-observability -> networking
network-policies -> networking
quickstart/<lang> -> quickstarts/languages
stateful-workload-filestore -> databases
terraform -> .github (these aren't GKE samples so I'm hiding it away)
try-gke -> demos
whereami -> quickstarts
windows-multi-arch ->  windows
wordpress-persistent-disks ->  quickstarts
workload-metrics -> observability
```

## How to review?
[...]

PR #1025, opened 2023-11-17
Remove sample duplicates

bourgeoisor:
This PR is a follow up to https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/987. In that previous PR, samples were duplicated to be moved in their respective new locations. This current PR removes the previous, now deprecated locations.

Before: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/tree/3d47e0c10115446c93c74ebcc5d5ff032e86ac2c

image

After: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/tree/issue-remove-sample-duplicates

image

## How to review?

Verify that all non-category directories are deleted, and in turn that category directories were not accidentally deleted. This is the original plan, which this PR should be following:
image


minherz (approved):
lgtm

Few clarification questions. Comparing the list of moved samples (in #987) to the list of deleted folders I see the `managements/migrating-node-pool/` folder does not appear in that list but is deleted. there is also deleted `migrating-node-pool/` folder. do you mean that both should be deleted?

bourgeoisor:
Good catch; that's an intentional corrective issue. The correct directory is expected by the tutorial (& the plan) to be `management` not `managements`, so I fixed that.

PR #1042, opened 2023-12-01
Replace missing "es" language with "fr"

vszal:
The loadgenerator here refers to a language key "es" which is not available per the model specifics: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/ai-ml/t5-model-serving/model/handler.py#L27

Update either here to "fr" or some other supported language, or add Spanish "es" to the model.


NimJay (approved):

NimJay:
Thanks for catching and fixing this. Approved.

NimJay:
All checks are passing. Merging.

PR #1048, opened 2023-12-07
fix: add new sample files for the dpv2 tutorial

PrajaktaW-Google:
Added two new yaml files for GKE 1.28 and later versions.  (https://critique.corp.google.com/cl/576505813)


bourgeoisor (changes requested):
Hi @PrajaktaW-Google!

- Don't forget licence headers for all files.
- It looks like your branch was based on an old version of the GKE Samples. don't forget to `git pull` the main branch before creating new work. In particular this PR works with the (incorrect) assumption that this sample is hosted in `network/` when in reality it's hosted in `networking/` (https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/tree/main/networking/dpv2-observability)

Issue #1056, opened 2023-12-15
whereami missing requests / limits


theemadnes:
need to specify some basic resource details in the deployment specs

PR #1057, opened 2023-12-15
added resource requests and limits for whereami

theemadnes:
related to https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/1056

otherwise, running on GKE Autopilot consumes too much memory

### DIFF
--- a/quickstarts/whereami/README.md
+++ b/quickstarts/whereami/README.md
@@ -108,10 +108,10 @@ spec:
         resources:
           requests:
             memory: "512Mi"
-            cpu: "500m"
+            cpu: "250m"
           limits:
             memory: "512Mi"
-            cpu: "500m"
+            cpu: "250m"
         ports:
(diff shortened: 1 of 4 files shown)


theemadnes:
@gbrayut @bourgeoisor

gbrayut:
Autopilot already defaults to 0.5cpu, so would be better to do 0.25 since that is currently the [minimal value](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#compute-class-min-max)

theemadnes:
> Autopilot already defaults to 0.5cpu, so would be better to do 0.25 since that is currently the [minimal value](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#compute-class-min-max)

done

bourgeoisor (approved):
LGTM

PR #1071, opened 2023-12-22
Issue 1006 add support for local Ipv6

theemadnes:
a bit complex because using `::` works for Flask but doesn't for gRPC, so needed to update configMap to suggest `[::]` and then strip the brackets in `app.py` when running in HTTP/Flask mode.


theemadnes:
also bumped Python version to `3.12.1`.

bourgeoisor (approved):
Following two modified file is missing a copyright header:
- quickstarts/whereami/app.py
- quickstarts/whereami/helm-chart/values.yaml

Besides that LGTM

PR #1083, opened 2024-01-03
feat: samples for upcoming Qdrant: Uploading docs guide

ganochenkodg:
Qdrant manifests and python code


databases/qdrant/docker/chatbot/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.11-slim-bookworm
```
bourgeoisor, line 1: Could you add to this file a step that simply builds the Dockerfile (just a `docker build`) so we have this added check?

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/.github/workflows/databases-qdrant-ci.yml

bourgeoisor: Also note that Renovate will immediately want to change this Python image version to 3.12

databases/qdrant/docker/embed-docs/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.11-slim-bookworm
```
bourgeoisor, line 1: Same comment: Could you add to this file a step that simply builds the Dockerfile (just a `docker build`) so we have this added check?

https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/.github/workflows/databases-qdrant-ci.yml

databases/qdrant/docker/chatbot/Dockerfile
ganochenkodg: As i remember some used here python libraries required to use python 3.11, can't say which ones right now

bourgeoisor: Required exactly 3.11, or "at least 3.11"?
In either case that's one more reason why I'm suggesting at minimum a "docker build" workflow so that we have at least that check having to pass if Renovate decides to bump any versions

databases/qdrant/docker/embed-docs/Dockerfile
BRV158: databases/qdrant/docker/embed-docs
databases/qdrant/docker/chatbot
ci steps added

bourgeoisor: Thank you!

bourgeoisor (approved):
LGTM, let me know when it's ready to be merged

BRV158:
Yep, it can be merged.

PR #1091, opened 2024-01-10
fix: Remove incorrect use of PORT env in Dockerfiles for languages samples

bourgeoisor:
b/314683443


NimJay (approved):
Looks good to me.

**Question:** Should we also update the use of $PORT in the [PHP sample's Dockerfile](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/HEAD/quickstarts/languages/php/Dockerfile#L24)?

bourgeoisor:
This PR has been lingering for a while, closing for now

PR #1122, opened 2024-01-30
Finetune code and fixes

ganochenkodg:
- Changed disk type in terraform modules
- Fixed vectors graph in the dashboard
- Updated python versions 3.11 -> 3.12
- Update langchain libraries and code to the actual versions


bourgeoisor (approved):

ganochenkodg:
@bourgeoisor i think you can, in case of some changes we will create an additional small PR, and now it's important to fix the code (langchain update issues)

PR #1148, opened 2024-02-21
Add Gemma open model serving engine examples

kenthua:
Gemma open model serving engine examles
- saxml
- vllm
- tgi
- trtllm

Gradio chat example


qchong:
@kenthua Can you please add region tags around the key parts of the sample code that we want to expose in the docs? The syntax is like this:

// [START region_tag_label]
...your code goes here...
// [END region_tag_label]

For a concrete example, you can refer to this sample (https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/HEAD/databases/gke-stateful-postgres/terraform/gke-autopilot/main.tf) which we embed in this doc (https://cloud.google.com/kubernetes-engine/docs/tutorials/stateful-workloads/postgresql).

LMK if you have questions about this request.

bourgeoisor (approved):
LGTM! I can merge any time.

PR #1177, opened 2024-03-05
Pgvector RAG guide

ganochenkodg:
YAML manifests, terraform and python code for the guide.


bourgeoisor (approved):

bourgeoisor (changes requested):
Oh, could you add a workflow to test that the two Dockerfiles can build, and that the Terraform scripts can validate?

You can find some example, e.g.: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/.github/workflows/cost-optimization-gke-vpa-recommendations-ci.yml

PR #1201, opened 2024-03-14
Samples refactoring: Chatbot universal guide

ganochenkodg:
TF and yaml code for unified RAG app guide


bourgeoisor:
@ganochenkodg this PR deletes a handful of files. Are we sure they're not used in any tutorials right now?

ganochenkodg:
@bourgeoisor it affects only chatbot guides, and no one of those were published at the moment. But yep, let's convert it to draft and wait for the final review

bourgeoisor:
@ganochenkodg feel free to ping me when the PR is stable and ready for review.

I assume there's also going to be changes in the tutorial wording as well? Is there an internal changelist up? They'll probably have to be merged around the same time so nothing breaks.

ganochenkodg:
@bourgeoisor Hi Olivier, you can check it. The guide was technically reviewed by Ameenah, as i know, and nothing will be changed hopefully. Also it can be easily merged without any breaks because it contains only the code for non-published guides (the one shared guide instead of 3 separate for different DBs).

bourgeoisor (approved):

PR #1212, opened 2024-03-21
Use jupyter notebooks

ganochenkodg:
Vector databases basic guides: use Jupyter notebooks instead of running scripts in the python shell


bourgeoisor (commented):
LGTM but let's not merge until we can also merge CL 622854406. Else, the doc page is going to be pointing to samples that don't exist anymore.

ganochenkodg:
>LGTM but let's not merge until we can also merge CL 622854406. Else, the doc page is going to be pointing to samples that don't exist anymore.

Yep, it affects the current Qdrant guide, lets wait 👍

bourgeoisor (approved):
Snippet bot looks good now, LGTM!

PR #1296, opened 2024-05-21
Add Gemma finetune on GPU single node code, dockerfile and kubernetes manifest

moficodes:
## Description

This PR add the python code, Dockerfile and Kubernetes manifests for Finetuning Gemma on GKE with multiple GPUs on a single Node.

## Tasks

* [x] Workflow files have been added / modified, if applicable.
* [x] Region tags have been properly added, if new samples.
* [x] All dependencies are set to up-to-date versions, as applicable.


moficodes:
@bourgeoisor

ai-ml/llm-finetuning-gemma/transformers/finetune.py
```diff
@@ -0,0 +1,237 @@
+# [START gke_aiml_llm_finetune_gemma_single_node_py]
```
bourgeoisor, line 1: Missing Google's copyright header here

moficodes: good catch. fixed.

ai-ml/llm-finetuning-gemma/transformers/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# [START gke_aiml_llm_finetune_gemma_single_node_docker]
+FROM nvidia/cuda:12.2.0-runtime-ubuntu22.04
+
+RUN apt-get update && \
+    apt-get -y --no-install-recommends install python3-dev gcc python3-pip git && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip3 install --no-cache-dir accelerate bitsandbytes datasets transformers peft trl torch
```
bourgeoisor, line 8: In general it's preferable to check-in a requirements.txt file with pinned versions and let Renovate bump these versions, instead of doing it in the Dockerfile directly. The reason why is because with this method, the Dockerfile is not deterministic anymore; if I build this image today, vs. if I build this image in a few weeks, it'll create two completely different images (even if the content of the Dockerfile hasn't changed).

ai-ml/llm-finetuning-gemma/transformers/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# [START gke_aiml_llm_finetune_gemma_single_node_docker]
+FROM nvidia/cuda:12.2.0-runtime-ubuntu22.04
```
bourgeoisor, line 2: It'd also check-in a workflow that does at minimum a `docker build` so we get a immediate heads-up if there's e.g. a dependencies conflict that prevents the image from building (especially if we're hardcoding packages not pinned to any versions)

You could just copy-paste this file and replace the pointers (name of the workflow, paths, etc) to reflect your sample: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/.github/workflows/ai-ml-gke-a100-jax-ci.yml

bourgeoisor (changes requested):
Just two small things, the rest looks good

ai-ml/llm-finetuning-gemma/transformers/Dockerfile
moficodes: Thats fair. Might just add versions of dependency in the Dockerfile. Otherwise its one more file for people to copy.

WDYT?

bourgeoisor (approved):
LGTM!

PR #1548, opened 2024-12-03
fix: Fix "helm push" of whereami major tag release

NimJay:
## Description

* For context, please see https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/1540.
* This change fixes the following error encountered in the Cloud Build trigger that runs against commits on the main branch:
```
Error: whereami-1.tgz: no such file
```
* See [build logs here](https://console.cloud.google.com/cloud-build/builds/9b5ea98e-ba39-4879-a1cd-1accd376cadd).

PR #1540, opened 2024-11-25
Introduce v1 tag for whereami image

NimJay:
## Description

* The `whereami` image is available at:
    * `us-docker.pkg.dev/google-samples/containers/gke/whereami`
    * ``
    * ``

This PR....

## Tasks

* [ ] The [contributing guide](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/.github/CONTRIBUTING.md) has been read and followed.
* [ ] The samples added / modified have been fully tested.
* [ ] Workflow files have been added / modified, if applicable.
* [ ] Region tags have been properly added, if new samples.
* [ ] All dependencies are set to up-to-date versions, as applicable.
* [ ] Merge this pull-request for me once it is approved.

PR #1550, opened 2024-12-04
Remove logic for pushing major-versioned whereami Helm chart

NimJay:
**Background**

This pull-request is a follow up to:
1. https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/1540 👈  This introduced the `v1` tag. But we saw errors in the building of the `v1` Helm chart.
2. https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/1548 👈  This led to a new error. See [Cloud Build trigger logs here](https://cloud.console.google.com/cloud-build/builds/bb167225-6223-495e-b55e-79f220fbdf64).

In the first pull-request, we introduced a `v1` tag for the `whereami`, that lets users pull the latest `1.y.z` image. In other words, docs could reference `wheream:v1` instead of `wheream:v1.2.3` so that those docs don't need to be updated every time there's a new `whereami` patch or minor release.

**The issue**

Unfortunately, unlike Docker container images, the use of a `v1` tag doesn't work for Helm charts/packages images. Helm version tags must always contain the major, minor, _and_ patch number. See [Helm docs](https://helm.sh/docs/topics/charts/#charts-and-versioning):

> Every chart must have a version number. A version must follow the [SemVer 2](https://semver.org/spec/v2.0.0.html) standard.

The `helm install` command already takes care of this for us. See [`helm install` docs](https://helm.sh/docs/helm/helm_install/#options):

> `--version string` - specify a version constraint for the chart version to use. This constraint can be a specific tag (e.g. 1.1.1) or it may reference a valid range (e.g. ^2.0.0). If this is not specified, the latest version is used

**The solution**

This pull-request removes the logic that tries to build a Helm chart/package versioned with just the major tag (`v1`).

### DIFF
--- a/quickstarts/whereami/cloudbuild.yaml
+++ b/quickstarts/whereami/cloudbuild.yaml
@@ -49,7 +49,7 @@ steps:
     helm package . # This creates a file similar to whereami-X.Y.Z.tgz
     helm push whereami-1.2.23.tgz oci://us-docker.pkg.dev/google-samples/charts
     # Push Helm chart which could be referred to with just the major version tag
-    helm push oci://us-docker.pkg.dev/google-samples/charts/whereami --version 1
+    helm push us-docker.pkg.dev/google-samples/charts/whereami:1.2.23
 
 images:
   - 'gcr.io/google-samples/whereami:v1'
--- a/quickstarts/whereami/cloudbuild.yaml
+++ b/quickstarts/whereami/cloudbuild.yaml
@@ -49,7 +49,7 @@ steps:
     helm package . # This creates a file similar to whereami-X.Y.Z.tgz
     helm push whereami-1.2.23.tgz oci://us-docker.pkg.dev/google-samples/charts
     # Push Helm chart which could be referred to with just the major version tag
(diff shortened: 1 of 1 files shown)


quickstarts/whereami/cloudbuild.yaml
```diff
@@ -48,8 +48,8 @@ steps:
     cd quickstarts/whereami/helm-chart
     helm package . # This creates a file similar to whereami-X.Y.Z.tgz
     helm push whereami-1.2.23.tgz oci://us-docker.pkg.dev/google-samples/charts
-    cp whereami-1.2.23.tgz whereami-1.tgz
-    helm push whereami-1.tgz oci://us-docker.pkg.dev/google-samples/charts
+    # Push Helm chart which could be referred to with just the major version tag
+    helm push us-docker.pkg.dev/google-samples/charts/whereami:1.2.23
```
NimJay, line 52:
```suggestion
    helm push oci://us-docker.pkg.dev/google-samples/charts/whereami:1.2.23
```

bourgeoisor (approved):
LGTM, leveraging Helm's version range instead makes sense.

Issue #1581, opened 2025-01-03
Serve an LLM with multiple GPUs in GKE is doesn't work and fails with The node was low on resource: ephemeral-storage.


raushan2016:
- **Impacted sample:** https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/main/ai-ml/llm-multiple-gpus/llama3-70b/text-generation-inference.yaml 
- **Impacted public tutorial:** https://cloud.google.com/kubernetes-engine/docs/tutorials/serve-multiple-gpu#autopilot
- The tutorial doesn't work end to end and the container keep on crashing with "The node was low on resource: ephemeral-storage. Threshold quantity: 10120387530, available: 9166736Ki. Container llm was using 57900024Ki, request is 0, has larger consumption of ephemeral-storage"
- Container logs is stuck at 13/30 
 ```
2025-01-03 11:06:50.367 PST
[2m2025-01-03T19:06:50.366843Z[0m [32m INFO[0m [2mtext_generation_launcher[0m[2m:[0m Download: [13/30] -- ETA: 0:06:42.769236
```
- Somehow instead of **/data** some other mount **/etc/hosts** is getting filled and eventually runs out of disk space.
```
root@llm-689555d8bf-62gjd:/etc# df
\Filesystem     1K-blocks     Used Available Use% Mounted on
overlay         98831908 75370476  23445048  77% /
tmpfs              65536        0     65536   0% /dev
/dev/nvme0n2   153707984       28 153691572   1% /data
tmpfs           62914560       12  62914548   1% /dev/shm
/dev/nvme0n1p1  98831908 75370476  23445048  77% /etc/hosts
tmpfs           62914560       12  62914548   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs           49441048        0  49441048   0% /proc/acpi
tmpfs           49441048        0  49441048   0% /proc/scsi
tmpfs           49441048        0  49441048   0% /sys/firmware
```
- The suspected change in the underlying GIT repo is https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/1495
- Using the old image `ghcr.io/huggingface/text-generation-inference:2.0.4` result in to successful run. Below is the disk usage for the same.
```
root@llm-fb5d99cb-569b7:/usr/src# df  
Filesystem     1K-blocks      Used Available Use% Mounted on
overlay         98831908  25463724  73351800  26% /
[...]

PR #1591, opened 2025-01-15
Reverting the TGI image version for LLAMA multiple GPUs in GKE samples

raushan2016:
## Description

The current image override the HF_HOME to /tmp from /data. Even after changing the mountpath to /tmp there is some regression in the newer TGI image which results into out of GPU memory on L4 and requires atleast A2 node. Rolling back the image version to get the sample working will investigation happen in the background.

Issue: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/1581

## Tasks

* [ ] Workflow files have been added / modified, if applicable.
* [ ] Region tags have been properly added, if new samples.
* [ ] Editable variables have been used, where applicable.
* [ ] All dependencies are set to up-to-date versions, as applicable.
* [x] Merge this pull-request for me once it is approved.

### DIFF
--- a/ai-ml/llm-multiple-gpus/falcon-40b/text-generation-inference.yaml
+++ b/ai-ml/llm-multiple-gpus/falcon-40b/text-generation-inference.yaml
@@ -29,7 +29,7 @@ spec:
     spec:
       containers:
       - name: llm
-        image: ghcr.io/huggingface/text-generation-inference:1.4.3
+        image: us-docker.pkg.dev/deeplearning-platform-release/gcr.io/huggingface-text-generation-inference-cu121.1-4.ubuntu2204.py310
         resources:
           requests:
             cpu: "10"
--- a/ai-ml/llm-multiple-gpus/falcon-40b/text-generation-inference.yaml
+++ b/ai-ml/llm-multiple-gpus/falcon-40b/text-generation-inference.yaml
@@ -51,7 +51,7 @@ spec:
         volumeMounts:
           - mountPath: /dev/shm
             name: dshm
-          # mountPath is set to /data as it's the path where the HF_HOME environment
+          # mountPath is set to /data as it's the path where the HUGGINGFACE_HUB_CACHE environment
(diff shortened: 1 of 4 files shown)


moficodes:
@kenthua I am ok with this change.

moficodes (approved):

kenthua:
In syncing with @raushan2016 , we have deeplearning container image that does work for this example.

kenthua:
can we add the `20241210-1804-rc0` image tag?

ai-ml/llm-multiple-gpus/falcon-40b/text-generation-inference.yaml
```diff
@@ -51,6 +51,9 @@ spec:
         volumeMounts:
           - mountPath: /dev/shm
             name: dshm
+          # mountPath is set to /data as it's the path where the HF_HOME environment
```
alvarobartt, line 54: For images lower than TGI 2.3 the environment variable was `HUGGINGFACE_HUB_CACHE` instead of `HF_HOME`

```suggestion
          # mountPath is set to /data as it's the path where the HUGGINGFACE_HUB_CACHE environment
```

ai-ml/llm-multiple-gpus/llama3-70b/text-generation-inference.yaml
```diff
@@ -58,7 +58,10 @@ spec:
         volumeMounts:
           - mountPath: /dev/shm
             name: dshm
-          - mountPath: /data
+          # mountPath is set to /tmp as it's the path where the HF_HOME environment
```
alvarobartt, line 61:
```suggestion
          # mountPath is set to /tmp as it's the path where the HUGGINGFACE_HUB_CACHE environment
```

alvarobartt (approved):
Thanks for the PR, added some nits! Also I'll check the issues with both Falcon 40B and Llama 2 70B and come back with another PR on top once those are solved, thanks for your time again! 🤗

moficodes (approved):

PR #1776, opened 2025-08-25
chore: Add license header check

NimJay:
* This change migrates us away from https://github.com/apps/license-header-lint-gcf (which is deprecated, as of Aug 5, 2025).
* We now run a GitHub Action that runs `go run github.com/google/addlicense@v1.1.1`.
* This is inspired by https://github.com/google/oss-rebuild/blob/main/.github/workflows/license.yml.


glasnt:
per discussion with @shawnmjones : this PR now successfully shows there are licencing issues elsewhere in the repo. 

These should be corrected, either in this PR or another PR, then this PR rebased with those changes to ensure the checks run green. 

Merging this PR without having the new check running green means all other PRs will fail for the same reason (the header check is a full-repo scan, so it will affect all PRs)

PR #2031, opened 2026-04-20
Update Dockerfile to use static-debian12 base image

NimJay:
* See internal Buganizer: [b/503111624](http://b/503111624)
* This pull-request aims to shrink the base image used for the `gcr.io/google-samples/hello-app:1.0` image and the `gcr.io/google-samples/hello-app:2.0` image.
* Merging this pull-request will trigger a Google Cloud Build trigger that rebuilds the `containers/gke/hello-app` image. Once rebuilt, I can delete the old images (manually). See [related cloudbuild.yaml](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/blob/1355d98e9eeef6fe8c6f230e2f4b24b20ef5914a/quickstarts/hello-app/cloudbuild.yaml).
* CI (specifically, `quickstarts-hello-app-ci`) is passing! 👍

PR #2033, opened 2026-04-23
Update Go base image version in Dockerfile

NimJay:
See my comment from Apr 23 at [b/503111624](http://b/503111624).
This is a follow-up to https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/2031, which only cleared out some of the CVEs we wanted to clear out.

### DIFF
--- a/quickstarts/hello-app/Dockerfile
+++ b/quickstarts/hello-app/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # [START gke_quickstarts_hello_app_dockerfile]
-FROM golang:1.24.3 AS builder
+FROM golang:1.26.2 AS builder
 WORKDIR /app
 RUN go mod init hello-app
 COPY *.go ./


quickstarts/hello-app/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # [START gke_quickstarts_hello_app_dockerfile]
-FROM golang:1.24.3 AS builder
+FROM golang:1.26.2 AS builder
```
NimJay, line 16: * 1.26.2 seems to be the latest version based on my search at: https://hub.docker.com/_/golang/tags
* This is a minor bump (which should guarantee backwards compatibility).

NimJay:
This will bring the CVE count to 0:
Screenshot 2026-04-23 at 11 50 57 AM

iennae (approved):
Thank you!

PR #2070, opened 2026-05-18
Update NemoRL on GKE scripts.

yuyantingzero:
## Description

This PR updates script for NemoRL on GKE tutorial. 
- fixing syntax issues 
- remove hardcoded lustre config
- adjust gemma training parameters, so it ends in a few hours rather than weeks
- add node selector for gke head, so it does not land on gpu

## Tasks

* [ ] Workflow files have been added / modified, if applicable.
* [ ] Region tags have been properly added, if new samples.
* [x] Editable variables have been used, where applicable.
* [x] All dependencies are set to up-to-date versions, as applicable.
* [x] Merge this pull-request for me once it is approved.

### DIFF
--- a/ai-ml/nemo-rl-on-gke/nemoRL/templates/raycluster-cluster.yaml.j2
+++ b/ai-ml/nemo-rl-on-gke/nemoRL/templates/raycluster-cluster.yaml.j2
@@ -1,4 +1,4 @@
-{# Copyright 2026 Google LLC
+# Copyright 2026 Google LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


ai-ml/nemo-rl-on-gke/nemoRL/templates/raycluster-cluster.yaml.j2
```diff
@@ -10,7 +10,7 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License. #}
```
NimJay, line 13: **Nitpick (non-blocking):** The `{#` and `#}` in these .j2 files was intentional. We need the Apache license header on the .j2 files, but we don't need them in the .yaml files generated as part of the tutorial.

NimJay (approved):
Thanks for doing this. I trust that you have tested this appropriately. Merging!

